PR #11, opened 2020-11-08
some improvements

longemen3000:
1. while the system function is useful, is not extensible. a dispatch is added on $SAFT(vec_of_strings) so if you want to create a PCSAFT, you can do it directly via:
```julia
PCSAFT(["diethyl ether]")
```
2. added basic unitful support.
this is added via the function `normalize_units(val) `(copied from a package i have). the function returns the SI unit of a unitful quantity. for example, now it is possible to do:
```julia
using Unitful
    Cp = get_isobaric_heat_capacity(model, create_z(model, [0.5, 0.5]), 1u"bar", 25u"°C")
```


pw0908:
Hi Andrés, 
Thanks a lot for this! We've been thinking of doing similar things for quite some time!
However, I won't be merging these onto master yet for 2 reasons:
1. We have made some significant changes to OpenSAFT in one of our development branches (currently we're working on SAFTgammaMie); if you plan to make further changes, would you mind making them on that branch instead? Alternatively, we're about a week away from merging the development branch with master; it might be easier to make changes then?
2. Some of the changes you're implementing here we have already (attempted) to do so in the development branch. We would really appreciate to hear your thoughts on this!
Feel free to get in touch with us if you want to get more directly involved in OpenSAFT and what we have planned for the package!

PR #14, opened 2020-11-18
Test errors

longemen3000:
this was originally part of the Eos Optimizations branch (ongoing) but i want to push this on development because of some bugs on getproperties.jl, those bugs where catched by adding `test_methods.jl` to the test suite. there are some errors marked as broken that need to be addressed:

on `test_database.jl`
```julia
@test_broken filepath == dirname(pathof(OpenSAFT)) * "/../database/PCSAFT/data_PCSAFT_like.csv"

```
on `test_methods.jl`
```julia
@test_broken get_volume(test_system, 1E5, 298)[1] ≈ 5.9069454773905654e-5 rtol = 0.01 #returns incorrect value (3.011768338630139e-5)

@test_broken get_crit_pure(test_system)[1][1] ≈ 1 rtol = 1E10 #T_scale not defined

@test_broken get_speed_of_sound(test_system, 1E5, 298)[1] ≈ 1 rtol = 1E10 #requires that the model has Mr
```

Hopefully, in the EoSOptimizations branch, the last error could be fixed, but i don't know about the rest.

This branch also has a rework of the `system` function, exporting a similar one `System`, that returns the same inputs, but its totally decoupled, so new models could be added without modifying the function. for example, those two functions return the same:

```julia
 s1 = system(["Water"],"SRK")
 s2 = System(["Water"],"SRK")
```
but the second one also admits symbols and synonyms (only added VdW and SAFTγMie) :
```julia
 s3 = System(group_multiplicities,"SAFTγMie")
 s3 = System((["Water"],:PR)
```
in the ongoing work, im trying to port the EoS from a Dict-Style to an Array-Style, with the decoupling helping on that.

### DIFF
--- a/src/OpenSAFT.jl
+++ b/src/OpenSAFT.jl
@@ -34,8 +34,9 @@ include("models/eos/Cubic/CPA.jl")
 include("models/eos/eos.jl")
 
 include("models/system.jl")
+include("models/system2.jl")
 
-export system
+export system,System
 export eos,ideal
 
 using Unitful
--- a/src/methods/getproperties.jl
+++ b/src/methods/getproperties.jl
@@ -28,7 +28,7 @@ end
 function ∂2f(model,v,t,z)
     f(w) = eos(model,z,first(w),last(w))
     v,t = promote(v,t)
-    vt_vec =   SVector(_v,_t)
+    vt_vec =   SVector(v,t)
     ∂result = DiffResults.HessianResult(vt_vec)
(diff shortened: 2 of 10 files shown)


longemen3000:
ohhh, the test only fails on windows, i see, im gonna try to fix that

pw0908:
Shall I merge it then?

longemen3000:
i think i fixed the error, lets see if the CI agrees on that

PR #17, opened 2021-01-18
Minor fixes

dalum:
Hi there!  Nice work on this package!  I've had to adapt your code for a related project, and think I found two typos in the SAFT-gamma Mie implementation, as well as in the constants.  Not sure if you want to change them, if you're benchmarking against some other code with incorrect definitions as well, but wanted to give you a heads-up. :blush: There's a typo in `N_A`, and `k_B` was set to `1.380649e-23` by definition in 2019.

### DIFF
--- a/src/constants.jl
+++ b/src/constants.jl
@@ -1,5 +1,5 @@
-const N_A = 6.02214086e23 # mol-1
-const k_B = 1.38064852e-23 # m2 kg s-2 K-1
+const N_A = 6.02214076e23 # mol-1
+const k_B = 1.380649e-23 # m2 kg s-2 K-1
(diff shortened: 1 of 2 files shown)


pw0908:
Hi! Thanks a lot for this! We had been benchmarking primarily to pure-component figures from the original papers so it was unlikely we would have spotted either of these issues.

pw0908:
Out of curiosity, what are you using the code for?

dalum:
I work for a company called Hafnium Labs.  We're developing a larger system for thermodynamic property predictions, and I've been integrating it into that. :relaxed:

pw0908:
That's amazing to hear! Please let us know how it goes and if you need any assistance!

PR #28, opened 2021-08-01
Update methods.jl to fix molecular weights calculation

rameshputalapattu:
Fixing the molecular weight calculation. Updated the mapreduce function call with the correct order of operators, * and +


rameshputalapattu:
This pull request fixes issue #29

pw0908:
Hi rameshputalapattu, thanks a lot for finding this issue. I was having some problems with the speed of sound a while back and this might actually be responsible. 

I'll merge this right away.

PR #41, opened 2021-08-29
minor typo in the PT_flash function for the bubble point case

rameshputalapattu:
Hi @pw0908 
    Thank you for implementing PT_flash functionality. I was eagerly waiting for it.
    Looks like there was a minor typo which was causing the PT_flash function fail when encountering bubble point condition.
    Please see if this pull request is OK and would fix it.
Thanks and Regards
Ramesh Putalapattu


longemen3000:
Thanks!, it is really a work in progress at the moment, more updates will come in the following days, but I will merge anyway

PR #56, opened 2021-12-02
Add tp_flash & example notebook

denbigh:
This pull request contains a new file, src/methods/property_solvers/tp_flash/DifferentialEvolutiontp_flash.jl, in which a new function, tp_flash, is defined and exported. tp_flash solves the non-reactive flash problem via differential evolution.

I have also included an example notebook, examples/PTFlashDifferentialEvolution.ipynb, to demonstrate use of this function.

I am new to package development in julia, so I apologize if changes need to be made.

Feel free to free to use and modify this code in any way.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,13 +4,13 @@ authors = ["Hon Wa Yew <yewhonwa@gmail.com>", "Pierre Walker <pwalker@mit.edu>",
 version = "0.2.9"
 
 [deps]
-BlackBoxOptim = "a134a8b2-14d6-55f6-9291-3336d3ab0209"
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
 DiffResults = "163ba53b-c6d8-5494-b064-1a9d43ac40c5"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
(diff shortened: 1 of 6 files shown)


pw0908:
Thinking in terms of tests, what would say is the minimum required to ensure all aspects of your code works?

Would just testing for two-phase equilibrium be sufficient? Or should we test three-phase as well?

denbigh:
I think a two-phase test would be sufficient. There are a number of examples (with 2 and 3 species) in the example notebook. A three-phase simulation should work as well, but I believe you'd need >4 species to see a 3-phase region of phase space at fixed p and T.

pw0908:
There's a compatibility issue between BlackBoxOptim.jl and FillArrays.jl which we're trying to resolve. Really, we're just waiting for the BlackBoxOptim.jl folks to update their package...

pw0908:
I've now switched the optimiser package to Metaheuristic such that we don't have a compatibility issue. I've also added a simple test. Do you think that's sufficient?

longemen3000:
If the test passes, merge it, we can iterate over it later

pw0908:
Do you want to re-organise anything?

longemen3000:
Oh, yes, but if the functionality is working, the code reorganization can be done later

pw0908:
Alright, after 3 runs, I still can't get all tests to pass but I think Andrés' other update should be able to handle it.

denbigh:
Wonderful, thanks for making this process so smooth! I can see you have made my crude code more 'Julian' - I look forward to testing this to see if there are some performance improvements!

PR #72, opened 2022-03-19
Add chemical stability check

lucpaoli:
Added a chemical stability check. This returns a boolean distance based off of the unconstrained minimization of tangent plane distance function, TDP(w). If this is non-negative for any trial phase composition w, i.e. min(TDP) >= 0, then the chemical stability function returns true.

Issues:
This currently adds a new dependency of Optim.jl, which is likely unnecessary considering BlackBoxOptim is already present, as well as not constraining results to mole fractions in (0,1).
On top of this, I am suspect of the reliability of this and need to add both test cases, as well as validate the tangent plane distance function against a more traditional fugacity coefficient based form.

### DIFF
--- a/src/Clapeyron.jl
+++ b/src/Clapeyron.jl
@@ -12,8 +12,6 @@ using FillArrays: FillArrays
 using Roots: Roots
 using NLSolvers
 import BlackBoxOptim
-# For minimizing tangent plane distance function
-using Optim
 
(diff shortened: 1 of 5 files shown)


longemen3000:
@pw0908 is it okay to merge this and fix later? i can replace the Optim dependency with a NLSolvers solver

Issue #75, opened 2022-03-19
A question about defining a new component


kaige-web:
Hi! I'm trying to do some calculations about drugs using clapeyron.jl. I referred to the information in the user guide (using your own parameters). I have written:
model = PCSAFT(["griseofulvin","water"];userlocations=["C:/Users/GK/Desktop/Clapeyron.jl-master/PCSAFT_like.csv","C:/Users/GK/Desktop/Clapeyron.jl-master/PCSAFT_unlike.csv","C:/Users/GK/Desktop/Clapeyron.jl-master/PCSAFT_assoc.csv"])
T = 310.15
p = 1.01325e5
n10 = 0.001
n20 = 1
θ=[0.2 0.8]
x11=[n10*θ[1]/(n10*θ[1]+n20*θ[2])]
X1 = Clapeyron.FractionVector.(x11)
γ1=activity_coefficient.(model,p,T,X1)
![微信截图_20220319164149](https://user-images.githubusercontent.com/58410051/159114225-10b4f165-0f63-408e-86b4-bb5543f4705a.png)
![image](https://user-images.githubusercontent.com/58410051/159114278-7b809c2f-17bf-444d-9046-e766a2484ff3.png)
![image](https://user-images.githubusercontent.com/58410051/159114292-0d1dc1d8-f645-4894-bbdf-6c59e2bbf1ee.png)

However, when I calculated the activity coefficient of griseofulvin in water, the results were wrong. 
The activity coefficient of griseofulvin, 2.7570242281983007e-7, The activity coefficient of water 0.9999879858361557

In fact, the correct result should be 47384.9 for  griseofulvin and 1.04428e+10 for water.

I feel like I have a problem defining new parameters.  Could you provide an example of defining a new component?

pw0908:
Hi there; I'm currently away from my laptop and either Andrés or Paul may be able to check the code for you. But, just looking at your set up, you seem to have done everything correctly.

Thinking about the thermodynamics though, I'm a bit confused about the magnitude of the experimental data points? Water is almost pure in this system, thus, the activity coefficient should approach 1. Similarly, the interactions between water and griseofulvin are favourable so the activity coefficient should be less than 1. 

Did you mean to calculate the fugacity coefficient? Or perhaps did the experimentalist use another definition for the activity coefficients?

With regards to temperature dependent k parameters, you are correct; this has been a very long-standing problem in Clapeyron and we're trying to address it now that we are trying to implement parameter estimation.

[...]

PR #77, opened 2022-03-27
pharmaPCSAFT and combining rules

longemen3000:
this PR adds a new EoS and a new feature.
- `pharmaPCSAFT`: a PCSAFT variant, with temperature-dependent kij and and special handling for water if passed as `"water08"`.
- Combining rules for association: `AssocOptions` has a new keyword: `combining`, with options `sparse_nocombining` (default for all SAFT/CPA), `dense_nocombining` and `elliott` (default for `pharmaPCSAFT`).

addresses #75


longemen3000:
@pw0908 is the name ok?

longemen3000:
there was an error when calling for the temperature dependent sigma of water so the better results are just the work of the new combining rules

pw0908:
Looking at the code, it seems Delta assoc is using the wrong kappa? There shouldn't be a sigma in there. The units of kappa in pharmaPCSAFT are in angstrom

pw0908:
Ignore that, I realise the reason we can't replicate the activity is most-likely related to the association combining rules; Im good to merge for now

PR #91, opened 2022-05-13
Add helium to molarmass.csv

lucpaoli:
Add helium to molar mass, already defined in critical


longemen3000:
Weird, I thought it was already there haahah

PR #97, opened 2022-06-20
Reformulated chemical stability check to work as unconstrained minimisation

lucpaoli:
The main change is refactoring the bulk of chemical stability check into chemical_stability_analysis (which is not exported), where it's clearer that it's a p,T stability check rather than V,T. It also now works on mole numbers, allowing for unconstrained minimisation. chemical_stability still maintaines the previous API to be non-breaking.

One thing I'm unclear on in the chemical_stability_analysis is what tolerance should be defined for declaring a state unstable, e.g. whether a tdp of -1e-16 is a stable or unstable state.

In previous tests using Optim, I found BFGS outperformed Newton, though I'm unsure as to how to use BFGS with the defined Solver interface in Clapeyron. The stability check generally is quite slow, this could be improved by reducing the tolerance for convergence, as I do not believe the objective function for stability analysis is typically converged to machine precision (though feel free to correct me on that).

chemical_stability also checks for a metastable state, though this is currently just reported as "unstable".

Other smaller changes are defining hydrogen sulfide in WalkerIdea, as well as correcting an typo in Solvers.optimise with ``x0 isa number`` -> ``x0 isa Number`` (line 75)

Finally, changed diffusive_stability from calculating all eigenvectors and taking the minimum to just calculating the minimum eigenvector (line 51), for a minorly more efficient algorithm.

Once again, the large amount of changes are from autolinting. If this makes it unclear, I'm happy to redo PR.


longemen3000:
No problem, it happens.
One question, is it possible to reuse the code on `tpd.jl` for the stability?

lucpaoli:
Oh! Almost certainly, I hadn't noticed that file!

PR #98, opened 2022-06-25
Isofugacity calculation for bubble and dew points

gustavochm:
Hi!

I coded the -first version- of functions to compute dew and bubble points using the isofugacity method. All of these functions rely on a two step process:

1. The phase equilibria is solved by a combination of a newton's method to update the temperature and pressure and an inner successive substitution loop to update the phase composition. This part is usually robust and it is usually good when poor initial guesses are available. However, it can have a slow convergence rate.  
2. That's why the second part check if a solution was reached. If not, the functions will attempt to solve a multidimensional system of non-linear equations. This systems is solved by a second order converge method. The solver in this part usually requires a good initial guess, luckily we use the one obtained in step 1.

I think it could be useful to have a universal bubble and dew points solvers and input the method, as we did with the saturation point of the pure fluid. For the moment, I haven't modified any of the current bubble and dew functions 😅.

Also, probably these functions could benefit by using the same type on variables. I am not expert enough to code that yet though 🥲.

Please let me know what you think about these functions or if something about them is not clear.

Gustavo


longemen3000:
good!, same thing that the saturation methods, we need to create an API to integrate those, that call this internally.
Also, we have a backlog of PR ready to merge and we need to release a new version that contains the michelsen PR. in particular we need to merge first:
- https://github.com/ypaul21/Clapeyron.jl/pull/71 : new methods for saturation, bubble, dew temperature, new SuperAncilliary method for cubics (a fitted chebishev function that gives the correct saturation pressure for any PR, RK or vdW cubic within Float64 precision in 500 ns)
-  https://github.com/ypaul21/Clapeyron.jl/pull/99 : Cubic revamp and a lot of new cubics
After that, we need to release 0.3.7, and then we can start with this

longemen3000:
well, with the new release, we are now ready to start working on this. in general, i was thinking on something like `IsoFugacityEquilibria` and `ChemPotVEquilibria` methods that cover the four bubble/dew pressure/temperature combinations. there was also a discussion on adding a `non_condensable` api, so i need to think on how to do that. 
TODO list:

- [ ] make that api
- [ ] test

gustavochm:
Cool! I will also try to code something for non-condensables but it may take a while (I have a lot of uni stuff lately 🥲)

pw0908:
The non-condensable/non-volatile stuff is mainly something we want to try and do for all methods. This would simply involve setting one of the compositions for one of the components in the opposite phase to zero and eliminating the chemical potential equality related to that component. This is just something I've noticed which greatly speeds up some calculations and makes them more robust (e.g. gas+polymer solubilities).

gustavochm:
Hey, sorry about the forced push, my editor wasn't letting me sync for some reason. I just added a new file for the dew point calculation including the option for non-condensables (extra `non_condensables_list` option). I added them as new functions, but they basically do the same as the original dew point calculation functions but slicing to the components allowed in the liquid phase. 

I don't know if you prefer to keep these functions as a different method (like `IsoFugacityEquilibriaNonCondensable`) or keep the new ones and drop the original one. 

I will try to implement this for the bubble point calculation, but I'm still thinking about how to do it properly. Because if there is no non-condensable in the liquid phase, I don't know how the solver could choose between ignoring the non-condensable or adding it to the gas phase.

Gustavo

pw0908:
Thanks a lot for starting on this Gustavo!

For bubble points, it wouldn't be non-condensable species. It would be non-volatile. Same idea but the reverse logic. For tp-flash, both would need to be specified for VLE calcs.

gustavochm:
Hey!

I've working a bit into this.  I also added functions to compute bubble points that includes non-volatiles components ( extra `non_volatile_list ` argument). 

For the Michelsen TP flash, I realised that there is a bug in the `rr_vle_vapor_fraction(K, z)` function. For some reason it was not coverging correctly. I don't know the details of that solver so for the moment I just changed it back to Halley's method. I also created a new method for the TP flash (`ModifiedMichelsenTPFlash`) to allow non condensables and non volatiles components (extra `non_inx_list` and `non_iny_list` options). In theory this `non_in_list`  could be used for liquid-liquid and vapour-liquid equilibria. Also, as the original Michelsen TP flash this is based on successive substitution and minimisation of the Gibbs free energy. 

I did some test and its works really well 😋.

```
model = SAFTgammaMie(["hexane", "ethanol", "methane", "decane"])
T = 320.  # K
p = 1e5  # Pa
z = [0.25, 0.25, 0.25, 0.25]

x0 = [0.3, 0.3, 0., 0.4]
y0 = [0.2, 0.2, 0.6, 0.]

# normal TP flash
method = ModifiedMichelsenTPFlash(x0=x0, y0=y0, non_inx_list=[], non_iny_list=[], second_order=true)
X, nvals, G = tp_flash(model, p, T, z, method)
X
2×4 Matrix{Float64}:
 0.297101  0.282319  0.00235211  0.418227
 0.181118  0.202735  0.612165    0.0039813


# preventing decane to be in the vapour phase
method = ModifiedMichelsenTPFlash(x0=x0, y0=y0, non_inx_list=[], non_iny_list=[4], second_order=true)
X, nvals, G = tp_flash(model, p, T, z, method)
X
2×4 Matrix{Float64}:
 0.296763  0.281964  0.00236971  0.418904
 0.180785  0.202689  0.616526    0.0

# preventing methane to be in the liquid phase and decade to be in the vapour phase
method = ModifiedMichelsenTPFlash(x0=x0, y0=y0, non_inx_list=[3], non_iny_list=[4], second_order=true)
X, nvals, G = tp_flash(model, p, T, z, method)
X
2×4 Matrix{Float64}:
 0.297088  0.282261  0.0       0.420651
 0.181017  0.202738  0.616245  0.0

```

One thing though, is that this flash function still doesn't account for a correct model reduction. Mainly because if the model is reduced the index of the non volatiles/ non condensables may change e.e

Gustavo

pw0908:
Hi Gustavo,

Thank you so much for this! We really need to catch up and implement the non-condensable and non-volatile API in our own codes... it looks like such a convenient feature to use. I would be curious to see how one could implement it in our other flash algorithms.

Just to reassure you, we haven't forgotten about this PR. I just had to deal with a broken laptop and should be able to work on this shortly.

Best,

Pierre

longemen3000:
on the `index_reduction` / non-volatile clash, maybe we could benefit of a more generic `index_reduction` function that considers simultaneously empty/non-volatile/non-condensable/any other combinations

pw0908:
Just an idea but, if we change the `index_reduction` function to take in booleans instead, it could potentially be used for this purpose? In the case of `empty`, we'd just change the input to:
```julia
model_r,idx_r = index_reduction(model,z.>0.)
```
Or, we can just have two functions for `index_reduction`, one which accepts booleans and the other floats.

longemen3000:
> Just an idea but, if we change the `index_reduction` function to take in booleans instead

yes, i was thinking in just doing that (`index_reduction(model,z)` calls `index_reduction(z >. 0,model)`)

longemen3000:
some updates:
```
julia> Clapeyron.bubble_pressure(system,T,z,Clapeyron.FugBubblePressure())  #isofugacity bubble pressure, no initial points
(54532.237480651616, 7.78032376160735e-5, 0.045084612649139294, [0.5773403372123901, 0.42265966278760997])

julia> Clapeyron.bubble_pressure(system,T,z,Clapeyron.ChemPotBubblePressure()) #Clapeyron Default
(54532.249599971576, 7.78032376143955e-5, 0.04508461263835618, [0.5773403394700927, 0.42265966052990733])
```
there is a lot of boiler plate, but it seems necessary here. no other way. i will try to integrate the nonvolatile/noncondensable methods into one, via index reduction. but support of that in TPFlash seems harder, because we don't know a priori what type is a phase from outside `tp_flash_impl`.

longemen3000:
update:
all methods (`bubble_pressure`,`dew_temperature`,`dew_pressure`,`dew_temperature`) suport either `ChemPotX` or `FugX`, where X is the name of each method (`FugBubblePressure`, `ChemPotDewTemperature`). nonvolatiles/noncondensables support is baked in each struct, but it only does something in the `Fug` methods (by calling the `x_volatile`/`x_condensable` versions if the argument is present). the naming is kept intact in the functions, but on the struct, the keyword `nonvolatiles` is for non-volatiles and `noncondensables ` for non-condensables (those keywords are also used in the TP flash method)

The support for nonvolatiles/noncondensables is also baked in the default `MichelsenTPFlash` method. it will call the default method when those options aren't passed.

Also, all the `Fug` methods can use the default Clapeyron Initialization, that allows some methods to be called without any argument, but i found that the temperature versions are more sensitive to the initial point and outright fail with the default initialization.

with that, i would say that this PR is almost done for merging, we can add the support for nonvolatiles/noncondensables in `ChemPot` methods later

gustavochm:
Hey Andrés, thanks a lot for the enormous work to merge these functions!
Gustavo

PR #105, opened 2022-07-20
initial solid phase support, database revamp

longemen3000:
this PR does three things:
- starts support of models that could have a valid solid phase solution. this is done optionally by defining `x0_volume_solid(model,T,z)`. because of constant propagation, if a model does not define `x0_volume_solid`, it will not have any performance penalty related to searching on the solid branch. there is a new model `AnalyticalSLV` (doi:10.1023/a:1024015729095) that has a solid phase curve that can be calculated with the existing volume solver.
- database revamp: the new database infraestructure aims to be a little bit easier to understand and to reuse as much infraestructure as possible. some improvements:
  - we only read the files once. before we did two passes. one for the data, one for the sites.
  - there is less changes in data. the csv collected data is converted directly to a `ClapeyronParam` at the end
  - group reading uses the same functions as the other CSVtypes. so there is less code to read
  - if a database does not use separators, it is faster. if it does use separators, it is a little bit faster
  - AssocParams are built directly from the csv collected data instead of being made into a matrix of matrices and reconverted again into a list.
  - there is some initial support for `grouptype` disambiguation. the parsing is not done yet, but it is only a matter of adding a line of code that parses the group type from the csv.
  -  better verbose and error messages, because we now have more information available at the error site
- There is some backporting of some features from the `estimation` branches, namely
  - `getindex`, `setindex` on `SingleParam`, `PairParam`. for PairParam, the default `setindex!` is symmetric. you can opt-out   by doing `param[i,j,false] = 2` to only set the value at the specified indices
  - `copyto!` methods, to allow easier manipulation of params.
  - (indirect) deprecation of `model.icomponents`, now we use directly `model.components` as the length
  - (indirect) deprecation of `param.diagvalues`, we now build those with `diagvalues(param)`
  of course, because this is a non-breaking change, those fields are still there, but now they are unused.

What is missing: 
- porting the new `split_model` infraestructure: this will eliminate the need of using nested models on `SAFTgammaMie` and `UNIFAC` models
- porting the implace combining methods
In the solid part, some future work:
- we need `triple_point(model)`, `melting_pressure(model,T)`, `melting_temperature(model,p)`
- tentatively, maybe we could mix a solid eos with a fluid eos by using a piecewise approach?
```
function a_res(model::SolidAndFluid,V,T,z)
   lbv = lb_volume(model.fluid,z)
   if V < lbv
     return a_res(model.solidmodel,V,T,z)
   else
     return a_res(model.fluidmodel,V,T,z)
   end
end
```
there are some things that need to be considered there. we need some standarization to perform to allow those a_res to be compared, specially at equilibria calculations ( https://doi.org/10.1002/aic.15064)


pw0908:
Just a point of curiosity, since you're restructuring the parameters, could we add something for structure parameters? For s-SAFT-gamma Mie.

I think it would make sense to have something like this as we have something to handle vectors (SingleParam), matrices (PairParam) and 4-D tensors (AssocParams). Would make sense to have something for 3-D tensors.

PR #115, opened 2022-09-16
bounded RR and ASS in TP flash

gustavochm:
Hey guys,

I hope you are doing ok. I made some updates to the TP flash. I updated the RR solution procedure, to be bounded between 0 and 1 and  included acceleration into the SS solution procedure.

I also noticed the volume function was not using the analytical procedure for cubic EoS so I also changed that.

Another question that I have, is that when using cubic EoS for some reason by default the interaction parameters are obtained from the PC-SAFT database. Is that intentional?

Regards,
Gustavo

### DIFF
--- a/src/Clapeyron.jl
+++ b/src/Clapeyron.jl
@@ -63,6 +63,9 @@ include("utils/index_reduction.jl")
 #splitting models, useful for methods.
 include("utils/split_model.jl")
 
+# Gustavo: acceleration for successive substitution
+include("utils/acceleration_ss.jl")
+
 #Clapeyron methods (AD, property solvers, etc)
 include("methods/methods.jl")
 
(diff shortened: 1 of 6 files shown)


longemen3000:
SGTM  in my opinion, @pw0908 ?

pw0908:
Sorry for not replying soon enough; been working on some other stuff. Everything here looked great to me! Do we need to add new tests?

longemen3000:
Just some coverage ones, nothing too fancy

gustavochm:
I don't think so, the flash solvers are basically the same functions. The changed just made the RR solution procedure a bit more robust and the acceleration can save some iterations.

One thing though, regardless the interaction parameters for the cubic EoS. By default Clapeyron uses the kij from PC-SAFT, is that intentional?

pw0908:
Completely accidental; we'll fix that

Issue #85, opened 2022-04-19
Ammonia Multifluid


longemen3000:
https://doi.org/10.1016/j.fluid.2020.112496
https://doi.org/10.1016/j.fluid.2020.112749 (corrigendum)

Basically, a GERG2008-like model for fluids containing ammonia and related mixtures.

pw0908:
Interesting; seems like it should be quite quick to implement into Clapeyron?

It also covers a lot of funky species which is kinda neat!

ianhbell:
Yes, this model is not too complicated to implement, similar to existing multi-parameter empirical EOS, aside from the new term. It is implemented in teqp and CoolProp

ianhbell:
The EOS has been published: https://aip.scitation.org/doi/abs/10.1063/5.0128269

The SI includes implementation, and it is also already in teqp.

PR #147, opened 2023-03-10
Ammonia 2023 paper

longemen3000:
implements https://doi.org/10.1063/5.0128269. (#85)


src/utils/misc.jl
```diff
@@ -80,4 +80,12 @@ function doi2bib(doi::String)
         @warn "no internet connection"
         return ""
     end
+end
+
+function evalexppoly(x,n,v)
+    res = zero(x*first(n)*first(v))
+    for i in 1:length(n)
+        res += n[i]*x^v[i]
```
pw0908, line 88: I made a modification here since the previous implementation was giving an error

pw0908 (approved):

src/utils/misc.jl
longemen3000: Yeah, I saw, my mistake 😅

Issue #106, opened 2022-07-23
Split cubic & ideal EOS (and common backbone) into separate package(s)


stillyslalom:
The complexity of SAFT equations of state imposes quite a lot of overhead on this package, and that’s entirely justified for a research-grade EOS, but for simple day-to-day usage, it’d be great to have a low-overhead, fast-loading EOS package covering ideal/SRK/Peng-Robinson.

pw0908:
Splitting Clapeyron is an idea we've been thinking about for a while. This is mainly because some of the features we've been developing are getting quite niche and often warrant a package of their own. I'm still of the mindset that it's nice to have one unified package. However, as Clapeyron grows, it becomes more apparent that we should split it up.

That being said, I doubt SAFT represents a large overhead in Clapeyron? Surely something like our database parsers and methods should contribute much more. I might be wrong though.

longemen3000:
on limited testing:
```
julia> @time using Revise, Clapeyron #without any EoS
  8.001558 seconds (13.22 M allocations: 797.007 MiB, 5.96% gc time, 52.49% compilation time)

julia> @time using Revise, Clapeyron #with all EoS
  8.418290 seconds (13.28 M allocations: 802.694 MiB, 6.19% gc time, 50.00% compilation time)
```
So, the main overhead seems to be the code before the EoS (parsing and methods)

on how to split the package, the main future solution will almost surely be something like what they do in https://github.com/JuliaArrays/ArrayInterface.jl , but we still need to be the proper hierarchy of packages to load.

longemen3000:
with julia 1.8, the new macro `@time_imports` helps here:
```
julia> @time_imports using Clapeyron
      3.9 ms  StaticArraysCore
   1072.9 ms  StaticArrays
      1.9 ms  PackedVectorsOfVectors
    299.1 ms  FillArrays
      0.3 ms  CommonSolve
     57.6 ms  MacroTools
      2.0 ms  ConstructionBase
     56.6 ms  Setfield
     93.1 ms  Roots
      1.5 ms  PositiveFactorizations
     33.0 ms  NLSolvers
      4.0 ms  DataAPI
      0.6 ms  Compat
     17.4 ms  OrderedCollections
    112.6 ms  DataStructures
      0.7 ms  SortingAlgorithms
     15.0 ms  Missings
      9.5 ms  DocStringExtensions 57.54% compilation time
    162.9 ms  ChainRulesCore
[...]

PR #156, opened 2023-03-27
allow passing a `NamedTuple` to userlocations kwarg

longemen3000:
to make building models in the REPL even easier:
```
julia> model = PCSAFT(["a1"],userlocations = (;
        Mw = [1.],
        epsilon = [2.],
        sigma = [3.],
        segment = [4.],
        k = [0.0;;], #matrix
        n_H = [1],
        n_e = [1],
        epsilon_assoc = Dict((("a1","e"),("a1","H")) => 1000.),
        bondvol = Dict((("a1","e"),("a1","H")) => 0.001)))
PCSAFT{BasicIdeal} with 1 component:
 "a1"
Contains parameters: Mw, segment, sigma, epsilon, epsilon_assoc, bondvol
```
how it works?
when passing a named tuple, we perform a less strict parsing of the params, and transform to Single, Pair, Group or StructGroup depending on the shape of the data, or if the function is called from a `GroupParameter`. instead of doing locations -> csv -> tape -> param, we do namedtuple -> tape -> param and special-case when the component info is `nothing`

~~the big TODO here is the association parameters. at the moment, this works for any models that only uses single, pair or group parameters. but on association, this allows for only setting null association and it is hardcoded to `bondvol` and `epsilon_assoc`.~~
Association is done by passing `Dict{NTuple{2,Ntuple{2,String}},T}`. the special casing for `epsilon_assoc` and `bondvol` is still there to allow to set no-assoc models easily.

things to do:
- ~~tests~~
- docs

### DIFF
--- a/src/database/ParamOptions.jl
+++ b/src/database/ParamOptions.jl
@@ -10,8 +10,8 @@ const IGNORE_HEADERS = ["dipprnumber", "smiles", "cas"]
 """
     ParamOptions(;kwargs...)
 Struct containing all the options related to parameter parsing:
-* `userlocations::Vector{String} = String[]`: List of used-defined locations to search.
-* `group_userlocations::Vector{String} = String[]`: List of used-defined group locations to search.
+* `userlocations = String[]`: List of used-defined locations to search.
+* `group_userlocations = String[]`: List of used-defined group locations to search.
 * `asymmetricparams::Vector{String} = String[]`: List of pair parameters that follow that `param[i,j] ≠ param[j,i]`. if not set on asymmetric pairs, the asymmetric values will be overwritten!
 * `ignore_headers::Vector{String} =  ["dipprnumber", "smiles"]`: List of ignored headers.
 * `ignore_missing_singleparams::Vector{String} = String[]`: List of parameters where checking for missing parameter values (in `SingleParam`) or the diagonal (on `PairParam`) are ignored.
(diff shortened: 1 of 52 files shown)


pw0908:
I think we need to be careful as to how this is going to be used. It seems that this method only works for the SAFTs (and any EoS that doesn't depend on a sub-model). For a cubic, we may want to simply assemble it from Tc, Pc, and acentric_factor:
```julia
model = PR(["a1"],userlocations = (;
                      Mw = [15.],
                      Tc = [300.],
                      Pc = [1e6],
                      acentric_factor = [0.05]))
```
Those values will need to be communicated to the alpha model. I'd imagine this issue will also hold true for the activity coefficient models.

We can still merge this into master but with some strong caveats.

longemen3000:
> model = PR(["a1"],userlocations = (; Mw = [15.], Tc = [300.], Pc = [1e6], acentric_factor = [0.05]))
for submodels, we use the corresponding `x_userlocations`. in this case, `alpha_userlocations`
```
model = = PR(["a1"],userlocations = (; Mw = [15.], Tc = [300.], Pc = [1e6]), alpha_userlocations = (;, acentricfactor = [0.05]))
```

pw0908:
Oh, cool, perfect then

pw0908 (approved):

pw0908:
For clarity, shall we rename userlocations to userparams?

longemen3000:
it is possible, but the main reason i reused `userlocations` is that we don't add new keywords. for example, on a cubic, we would need to add:
- `userparams`
- `ideal_userparams`
- `alpha_userparams`
- `translation_userparams`
- `activity_userparams`
- `mixing_userparams`

longemen3000:
related issue: #106

pw0908:
I'm saying we rename userlocations and userparams, not an additional input

longemen3000:
I like the idea, but seems breaking, for 0.5 seems good

pw0908:
Lets open an issue and keep track of the things we want to add in v0.5. For now, we can merge this branch in.

Issue #144, opened 2023-02-07
Help - trying to make LLE with UNIFAC (Not an issue)


SalvadorBrandolin:
Hello, we are trying to perform a liquid-liquid equilibria flash with UNIFAC.
We have two problems:

The first one:

We perform the flash calculation with the default Clapeyron value for constants
and options and then we tried to evaluate the activity in each phase after the flash
calculation. We obtained different values...
```julia
using Clapeyron

model = UNIFAC(["water", "hexane"])

algorithm = MichelsenTPFlash(
    equilibrium = :lle, 
    K0 = [0.00001/0.99999, 0.99999/0.00001],
)

flash = tp_flash(model, 101325, 303.15, [0.5, 0.5], algorithm)

# Activity calculation on each phase

act_x = activity_coefficient(model, 101325, 303.15, flash[1][1,:]) .* flash[1][1,:]
act_y = activity_coefficient(model, 101325, 303.15, flash[1][2,:]) .* flash[1][2,:]
```

Output:
([0.9999900023219743, 0.062055849696312164], [0.0011464067122757193, 0.9999900003798906])

The second problem:

We tried to give to the MichelsenTPFlash the x0 and y0 values instead of K0 values.
And we obtained an error.
```julia

algorithm = MichelsenTPFlash(
    equilibrium = :lle, 
    x0 = [0.99999, 0.00001],
    y0 = [0.00001, 0.00009]
)

flash = tp_flash(model, 101325, 303.15, [0.5, 0.5], algorithm)

act_x = activity_coefficient(model, 101325, 303.15, flash[1][1,:]) .* flash[1][1,:]
act_y = activity_coefficient(model, 101325, 303.15, flash[1][2,:]) .* flash[1][2,:]

act_x, act_y
````

Output:

MethodError: reducing over an empty collection is not allowed; consider supplying `init` to the reducer

Stacktrace:
  [1] mapreduce_empty(#unused#::typeof(identity), op::Function, T::Type)
    @ Base ./reduce.jl:367
  [2] reduce_empty(op::Base.MappingRF{typeof(identity), typeof(min)}, #unused#::Type{Float64})
    @ Base ./reduce.jl:356
  [3] reduce_empty_iter
    @ ./reduce.jl:379 [inlined]
  [4] mapreduce_empty_iter(f::Function, op::Function, itr::Vector{Float64}, ItrEltype::Base.HasEltype)
    @ Base ./reduce.jl:375
[...]

PR #164, opened 2023-04-05
TP-Flash with activity models

longemen3000:
this PR enables the use of `tp_flash` with activity models in `MichelsenTPFlash`. there is support for non-volatiles/non-condensables and LLE. on VLE, performs the full approach for the calculation for K. (including Poynting Correction Factors).

what is missing at the moment is the optimization part, on LLE, i can just take the code for `LLE_point`,but code for VLE is missing.

solves #144 as originally intended. 

Because of the refactoring required to do this, `RRTPFlash` now supports the same (LLE,VLE,non-volatiles,non-condensables, and activities), because it now uses the same function to calculate those.


longemen3000:
this is ready for review/merge. the gibbs energy optimizer is pending, but rachford rice seems to handle the problem well (the solver was changed from a bracketed Halley to bracketed Chebyshev, see [Roots.jl/](https://github.com/JuliaMath/Roots.jl/issues/359). 

On UNIFAC, i did a second take on how to optimize it, this time using `thermopack` notes. the code changes made UNIFAC around 10x faster (and this time seem correct (they respect γi(xi = 0) -> 1).

A lot of the changes in other files was on removing `kwargs...` that were left from the multiple estimation branches. they cause problems because you could pass a mispelled keyword (`uselrocations`) and it will be silently dropped instead of explicitly erroring

pw0908 (approved):

PR #176, opened 2023-04-25
Add PPCSAFT implementation

pw0908:
Benchmarked and tested implementation of Polar-PCSAFT.


pw0908:
The test I've added is for a ternary mixture given the a_3 term involves 3 components. However, I couldn't find any examples online of a ternary mixture to benchmark against

longemen3000 (approved):

longemen3000:
the loops seems to be correct. i also made the `a_polar` contribution non-allocating

longemen3000:
Failing test is unrelated, good to merge

PR #181, opened 2023-05-17
Update SAFTgammaMie_groups.csv

Kelethabetse:


Kelethabetse:
modelling benzoic acid in ethanol, compound added to database

Issue #172, opened 2023-04-17
LinearAlgebra.SingularException in examples/multi-component_vle_vlle_lle_crit notebook


tkeskita:
Hi,

running this notebook stops at 3rd input cell with error:

LinearAlgebra.SingularException(5)

Stacktrace:
  [1] checknonsingular
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/factorization.jl:19 [inlined]
  [2] checknonsingular
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/factorization.jl:21 [inlined]
  [3] #lu!#136
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/lu.jl:85 [inlined]
  [4] #lu#140
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/lu.jl:273 [inlined]
  [5] lu (repeats 2 times)
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/lu.jl:272 [inlined]
  [6] \(A::Matrix{Float64}, B::Vector{Float64})
    @ LinearAlgebra /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:1136
  [7] default_newton_linsolve(d::Vector{Float64}, B::Matrix{Float64}, g::Vector{Float64})
    @ NLSolvers ~/.julia/packages/NLSolvers/lRxce/src/quasinewton/approximations/newton.jl:15
[...]

PR #185, opened 2023-06-22
Fix detection of real cubic roots

Liozou:
Thank you for this great package!

I encountered the following behaviour when looking at CO2 with a cubic equation of state:
```julia
julia> system = EPPR78(["carbon dioxide"]);

julia> Clapeyron.volume(system , 3311.0u"bar", 400.0u"K")
ERROR: UndefVarError: `vl` not defined
Stacktrace:
 [1] volume_impl(model::PR{…}, p::Float64, T::Float64, z::SVector{…}, phase::Symbol, threaded::Bool, vol0::Nothing)
   @ Clapeyron ~/.julia/packages/Clapeyron/JJGLA/src/models/cubic/equations.jl:269
 [2] volume(model::PR{…}, p::Float64, T::Float64, z::SVector{…}; phase::Symbol, threaded::Bool, vol0::Nothing)
   @ Clapeyron ~/.julia/packages/Clapeyron/JJGLA/src/methods/property_solvers/volume.jl:118 [inlined]
 [3] volume(model::PR{…}, p::Quantity{…}, T::Quantity{…}, z::SVector{…}; phase::Symbol, output::Unitful.FreeUnits{…})
   @ Clapeyron ~/.julia/packages/Clapeyron/JJGLA/src/methods/unitful_methods.jl:58 [inlined]
 [4] volume(model::PR{…}, p::Quantity{…}, T::Quantity{…}, z::SVector{…})
   @ Clapeyron ~/.julia/packages/Clapeyron/JJGLA/src/methods/unitful_methods.jl:55 [inlined]
 [5] top-level scope
   @ REPL[3]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

Arguably worse, for a slightly different pressure, there is no error but the answer is obviously wrong (the volume is negative):

```julia
julia> Clapeyron.volume(system , 3363.1u"bar", 400.0u"K")
-5.070066466695018e-5 m^3
```

After investigating, it turns out that this is due to a faulty heuristic used to determine which roots of a cubic polynomial are real. This PR implements a cleaner heuristic that checks the value of the polynomial between the real parts of the roots, and compares the sign of these values to determine more precisely whether the roots are real. This yields more satisfying answers:
```julia
julia> Clapeyron.volume(system , 3311.0u"bar", 400.0u"K")
3.363139140634349e-5 m^3

julia> Clapeyron.volume(system , 3363.1u"bar", 400.0u"K")
3.3545657072357283e-5 m^3
```

~And as a bonus, it seems that this bugfix incidentally makes the test for https://github.com/ClapeyronThermo/Clapeyron.jl/issues/172 pass (whereas it was marked as `@test_broken`).~

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Clapeyron"
 uuid = "7c7805af-46cc-48c9-995b-ed0ed2dc909a"
 authors = ["Hon Wa Yew <yewhonwa@gmail.com>", "Pierre Walker <pwalker@mit.edu>", "Andrés Riedemann <andres.riedemann@gmail.com>"]
-version = "0.4.11"
+version = "0.4.12"
 
 [deps]
 BlackBoxOptim = "a134a8b2-14d6-55f6-9291-3336d3ab0209"
--- a/src/models/cubic/equations.jl
+++ b/src/models/cubic/equations.jl
@@ -215,24 +215,14 @@ function crit_pure_tp(model::ABCCubicModel)
 end
 
 function volume_impl(model::ABCubicModel,p,T,z=SA[1.0],phase=:unknown,threaded=false,vol0=nothing)
-    lb_v   =lb_volume(model,z)
+    lb_v = lb_volume(model,z)
     if iszero(p)
         vl,_ = zero_pressure_impl(model,T,z)
         return vl
     end
     nRTp = sum(z)*R̄*T/p
     _poly,c̄ = cubic_poly(model,p,T,z)
-    sols = Solvers.roots3(_poly)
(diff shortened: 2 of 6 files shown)


Liozou:
Nevermind the "bonus", it seems that CI is a bit flaky considering the last runs. Should I erase the `@test_broken` into `@test` conversion?

longemen3000:
Just comment that test for now. Can you update the version number?, so we can release a new version as soon this is merged

Liozou:
Sure, I just did that (and added a small comment by the test to explain what the issue is)!

longemen3000 (approved):

PR #194, opened 2023-08-21
Solid solubility

pw0908:
Some preliminary work on a solid solubility solver. Right now we can calculate both the eutectic point and the solubility of a solid at a given temperature, pressure and solvent composition. This makes use of the CompositeModel API. A Simple example of this would be:
```julia
julia> model = CompositeModel(["water","ethanol","paracetamol"]; liquid=PCSAFT,gas=nothing,saturation=nothing,melting=nothing,solid=SolidSolubility,solid_userlocations=(;Hfus=[6000.,4969.00,27000], Tm=[273.15,158.70,443.6]))
Composite Model:
 Liquid Model: PCSAFT{BasicIdeal}("water", "ethanol", "paracetamol")
 Solid Model: SolidSolubility("water", "ethanol", "paracetamol")


julia> solid_solubility(model,1e5,298.15,[0.5,0.5,0.];solute=["paracetamol"])
3-element Vector{Float64}:
 0.4905731673763385
 0.4905731673763385
 0.01885366524732288
 ```

### CODE REVIEW

**Comment by pw0908** [comment]
Things that need to be done before merging:
[] Add tests
[] Add API docs
[] Add example notebooks?
Some feedback:
* The API is really flexibility because of CompositeModel. However, it is really intricate and not straight-forward. Could we create a sub-model of CompositeModel which handles solid solubility only where we'd only need to specify the liquid model?

**Comment by longemen3000** [comment]
> Could we create a sub-model of CompositeModel which handles solid solubility only where we'd only need to specify the liquid model?

You could create a function that does that, something ĺike this?
```julia
function Solid(components, liquidmodel = PC-SAFT, solidmodel =SolidSolubility, liquid_userlocations = String[], solid_userlocations = String[], verbose) 
li= init_model(liquid, components, liquid_userlocations, verbose)
so = init_model(solidmodel, components, solid_userlocations, verbose) 
return CompositeModel(components, liquid=li, gas=nothing,saturation=nothing,melting=nothing,solid=so) 
```

PR #204, opened 2023-09-19
Update cosmosac

pw0908:
Added functionalities where, if COSMO-SAC parameters are not in the Clapeyron database, they will be looked up in the NIST/COSMOSAC database.

@ianhbell do you have any opposition to this? We can add the relevant references to the model to ensure both your database and the folks who built the database receive the appropriate recognition.


ianhbell:
That makes sense to me. I would recommend a strong comment in the docs, perhaps a print statement in the console that you must manually silence, indicating the license terms within which the COSMO-SAC files are made available. They are not allowed to be used for all purposes.

pw0908:
Great! I've changed the implementation so that users must intentionally select the nist database to use it, at which point a warning is thrown to let them know to check the license.

The docs themselves are getting an overhaul on a separate branch where I've added similar warnings.

src/models/Activity/COSMOSAC/utils.jl
longemen3000, line 1: We could have this file as `COSMOSAC_Identifiers.csv`,already downloaded. In the long term, we should have the inchikey in the Identifiers.csv file. Anyway, seems good to merge

longemen3000 (approved):

src/models/Activity/COSMOSAC/utils.jl
pw0908: I've added this.

longemen3000:
test fail is unrelated

PR #217, opened 2023-10-19
Ppcsaft qq dq: Revision Request

beomseok-kang:
Hi. Until now,

Confirmed:
Pure component quadrupole-quadrupole interaction working.
All dipole-dipole interactions working.

Confirmed not working:
Mixed component quadrupole-quadrupole interaction

Not tested yet:
Any quadrupole-dipole interactions (although they are written in the codes already)


src/models/SAFT/PCSAFT/variants/PPCSAFT.jl
```diff
@@ -73,41 +80,191 @@ function a_res(model ::PPCSAFTModel, V, T, z)
...
-    dipole = model.params.dipole.values
-    if all(iszero,dipole)
-        return zero(V+T+first(z))
-    end
-    a₂ = @f(a_2,_data)
+    μ̄² = model.params.dipole2.values
+    Q̄² = model.params.quadrupole2.values
+    has_dp = !all(iszero, μ̄²)
+    has_qp = !all(iszero, Q̄²)
+    if !has_dp && !has_qp return 0. end
```
longemen3000, line 87: in general, i recommend trying to use a pattern like the following:
```
_0 = zero(V+T+first(z))
if !has_dp && !has_qp return _0  end
```
0.0 is a `Float64`, and that could correspond to the type of the value when using `eos(model,V,T,z)`, but for other properties, AD is involved, and that zero could actually be a `ForwardDiff.Dual` instead

src/models/SAFT/PCSAFT/variants/PPCSAFT.jl
```diff
@@ -143,20 +300,67 @@ function a_3(model ::PPCSAFTModel, V, T, z, _data=@f(data))
...
+    bij = _ppcsaft_corr_poly(consts.corr_b,m̄)
     cij = aij .+ bij .* ϵT
     return evalpoly(η,cij)
 end
 
-function J3(model, V, T, z, i, j, k, η = @f(ζ,3),m = model.params.segment.values)
-    corr_c = PPCSAFTconsts.corr_c
-    m̄ = min(cbrt(m[i]*m[j]*m[k]),2*one(η))
+function J3(model, V, T, z, type, i, j, k, η = @f(ζ,3),m = model.params.segment.values)
+    corr_c = eval(Symbol(type*"_consts")).corr_c
```
longemen3000, line 354: try to write this in a conditional instead, this is really bad in terms of performance. you could also do something like:
```
function get_ppcsaft_consts(x)
if x == :DD
  return DD_consts
elseif x ==:QQ
  return QQ_consts
else
  return DQ_consts #fallback, we always return something of the same type
end
end


```

src/models/SAFT/PCSAFT/variants/PPCSAFT.jl
```diff
@@ -73,41 +80,191 @@ function a_res(model ::PPCSAFTModel, V, T, z)
...
+    return _a_2
+end
+
+function a_3(model ::PPCSAFTModel, V, T, z, type, _data=@f(data))
+    dp_comps, qp_comps = @f(polar_comps)
+    P̄² = []
+    p_comps = []
+    p = 0
+    coeff = 0.
+    if type == "DD"
```
longemen3000, line 196: try using symbols (`:DD`) instead of strings `"DD"`. symbols don't allocate (at the cost of a slightly higher compilation time)

beomseok-kang:
Thank you!!

PR #218, opened 2023-10-20
Ppcsaft DQ: revision request

beomseok-kang:
DQ code revision request


pw0908:
As an aside, it appears that using the Quadrupolar (and quadrupolar-dipole) contributions isn't particularly common in the literature. As such, I suggest two modifications:
1. Define a new EoS, QPPCSAFT which uses the quadrupolar and dipole contributions and leave PPCSAFT on its own.
2. Have the use of the quadrupolar-dipolar term be optional as I've found that a few papers don't always use this term even when both quadrupoles and dipoles are present.

pw0908 (approved):

PR #241, opened 2023-12-14
Add Unitful support to all documented bulk properties

Liozou:
The documentation states (at the bottom of [Basic usages](https://clapeyronthermo.github.io/Clapeyron.jl/dev/user_guide/basic_usage/)):
```
Clapeyron also supports physical units through the use of Unitful.jl.
[...]
This is only supported for bulk properties.
```

This PR makes this necessary condition sufficient as well, by implementing unit support for all [documented Bulk properties](https://clapeyronthermo.github.io/Clapeyron.jl/dev/properties/bulk/).

I only added a couple of tests, I don't know if this requires being extensive and adding one test per method.

If you plan on extending unit support to all single phase and multiphase properties as well, it might be worth writing a macro to automate some of that work, because adding each method individually in ClapeyronUnitfulExt.jl is not really manageable. It also leads to brittle code, because it's easy to miss reflecting a change in the Unitful extension if a method is updated in Clapeyron.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -38,7 +38,7 @@
 # v0.5.3
 
 ## New Features
-- Databases were standarized according to CAS. almost all components present in Clapeyron.jl databases are present in `@DB/properties/identifiers.csv`.
+- Databases were standardized according to CAS. almost all components present in Clapeyron.jl databases are present in `@DB/properties/identifiers.csv`.
 - COSMOSAC-2002 (`COSMOSAC02`),COSMOSAC-2010 (`COSMOSAC10`) and COSMOSAC-dispersion (`COSMOSACdsp`) can now read files from the NIST database found at https://github.com/usnistgov/COSMOSAC . to use those parameters, pass the keyword `use_nist_database = true`
 - New model: doubly association perturbation theory (`DAPT`)
 - New model: PCSAFT with association dependent hard sphere diameter (`ADPCSAFT`)
--- a/docs/src/properties/bulk.md
+++ b/docs/src/properties/bulk.md
@@ -26,20 +26,20 @@ Clapeyron.pip
 
 In general almost all bulk properties follow the pattern:
 ```julia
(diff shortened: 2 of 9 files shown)


longemen3000:
Nice!

> If you plan on extending unit support to all single phase and multiphase properties as well, it might be worth writing a macro to automate some of that work

It is an idea worth exploring. something like this comes to mind:
```
@defineunits T,p,V = crit_pure(model)
```
but for the moment, this PR is good to go (given that the tests pass)

longemen3000 (approved):

Liozou:
Great, thanks for merging!

Regarding the macro, I was thinking more of adding it before every method definition in Clapeyron that implicitly assumes a unit, and for which another version should be provided for the Unitful extension. For instance, you would write something like:
```julia
@unitsupport function x0_volume_liquid(model::SAFTVRMieModel, T::"K", z)::"m^3"
    v_lb = lb_volume(model,z)
    return v_lb*1.5
end
```
in Clapeyron, and this would rewrite the signature by removing the `::"unit"` parts to make the code legal, while also `push!`ing the signature `x0_volume_liquid(model::SAFTVRMieModel, T::"K", z)::"m^3"` to a global variable, say `const UNITSUPPORT = Expr[]`. Then, in `ClapeyronUnitfulExt.jl` you would collect the signatures from `C.UNITSUPPORT` and make a new method, for instance in that case it should create:

```julia
function C.x0_volume_liquid(model::($modeltype), T::Unitful.Temperature, z; output=u"m^3")
    _T = ustrip(u"K", T)
    res = C.x0_volume_liquid(model, _T, z)*u"m^3"
    return uconvert(output, res)
end
```

Such a design would allow to specify everything directly in the initial code, so there would be no duplication of information, and no risk of forgetting to update the extension at the same time as the code. If I manage to find some time during winter break, I'll try to see if that's doable (if you find the design OK).

PR #245, opened 2024-01-02
Remove Unitful from [deps]

Liozou:
`Unitful` is a weak dependency, not a strong one (it could be both, as in https://github.com/JuliaLang/julia/issues/50028, but that's not the case here), so it should not be in the [deps] section of the Project.toml.


longemen3000:
Unitful was moved from a required dependency to a weak one. In Julia 1.9 onwards, dependencies in [deps] are ignored if they are also in [weakdeps], but for earlier Julia versions (in particular 1.6) it is required.

Liozou:
Silly me, I forgot that, thanks for clarifying!

PR #247, opened 2024-01-04
Add vol0 keyword argument to bulk properties

Liozou:
This PR documents the `vol0` keyword argument of `volume`. It also adds that same keyword argument to all bulk properties that implicitly call `volume` in order to pass it along, similar to `threaded` and `phase`.

I also added a test that fails when the `vol0` keyword is not given, to illustrate a typical use-case.


longemen3000 (approved):
Good catch!

PR #248, opened 2024-01-04
Improve PR Ω constants 

longemen3000:
This PR changes the values of the Ωa and Ωb constants for the Peng-Robinson equation of state,to be accurate within `Float64` precision. Most of the pull request is just fixing the tests. Constants are taken directly from the cubic superancillary paper (https://doi.org/10.1021/acs.iecr.1c00847)


longemen3000:
Note that RK and vdW EoS don't suffer this problem, only PR has imprecise constants

PR #272, opened 2024-05-04
MultiPhase TP-Flash

longemen3000:
this PR adds a new multiphase tp-flash.
it is based on the michelsen multiphase method (tpd + RR + gibbs minimization) with some modifications, listed below:
- the RR calculation of phase fractions was based on a paper by Okuno et al. (2010), it has a good definition of a guarded newton step that allows easy convergence.
- the gibbs minimization was done in lnK-β-v space (volume solver) instead of mole numbers.
- the creation of new phases is done by using the result of the stability calculation to create a new phase, instead of using wilson k values. if we don't have critical points, this is faster.

This PR also speeds up `tpd` significantly, to the point it is safe to call it multiple times.
supercedes #108


pw0908:
Have you tried the seven-phase system?

longemen3000:
> Have you tried the seven-phase system?

Not yet!, building one in a moment

pw0908:
I can get four phases to appear:
```julia
model = SRK(["hexane","aniline","water","nitrogen"];mixing=MHV2Rule, activity=UNIFAC)
z = ones(length(model))/length(model)

(x,n,G) = tp_flash(model,1e5,298.15,z,MultiPhaseTPFlash(nacc=20))
```
I'll try to add a fifth with a halo-alkane

longemen3000:
Ideally, the add_phases/replace_phases could be replaced by a HELD/GELD style loop, that option would be faster than the successive tpds, I think

On the todo list, one can, also reformulate the Gibbs minimization in terms of mass logK-β-v to allow it to solve dissimilar systems

pw0908:
Once electrolytes are done, Ill come back around to HELD...

Did you manage to do the mass version for regular RR?

pw0908:
Also, for five phases, it manages to realise there should be five phases, but it doesn't converge:
```julia
model = SRK(["hexane","aniline","1,1,2-trichloroethane","water","nitrogen"];mixing=MHV2Rule, activity=UNIFAC)
z = ones(length(model))/length(model)

(x,n,G) = tp_flash(model,1e5,298.15,z,MultiPhaseTPFlash())
x
```

longemen3000:
Some ideas:
- calculate tpd over mixture of phases instead of each phase
- change RR by a full SS system in the case n phases >2

Project.toml
```diff
@@ -57,6 +57,7 @@ ForwardDiff = "^0.10"
 JSON3 = "1"
 LinearAlgebra = "1"
 LogExpFunctions = "^0.3.24"
+Metaheuristics = "2, 3"
```
pw0908, line 60: Wait, why did we add Metaheuristics as a dependency?

longemen3000: It is just as compat, it only applies when the extension is loaded

pw0908 (approved):

PR #296, opened 2024-09-22
Hanna

se-schmitt:
Add HANNA model (https://arxiv.org/abs/2407.18011).

Open issues:

- [ ] Registration of `ClapeyronHANNA.jl`
- [ ] Add `ClapeyronHANNA.jl` to docs project
- [ ] "Register" CI for ClapeyronHANNA


longemen3000 (approved):
seems good to me, the only thing missing is a valid `LICENSE` file, on the CI infrastructure, i think we can just integrate it with Clapeyron itself, so it requires changes on our part. (probably a conditional action)

pw0908 (approved):

pw0908:
Hi Sebastian, this is amazing! Thanks so much for taking the time to implement this. I've updated some of the code so that users don't always have to specify smiles. Just a few questions:
1. Is there a reason why you've defaulted to using the BasicIdeal model? If you want to be able to do VLE calculations, it would be easier to specify PR as a default (which is what we do in most activity coefficient models).
2. Could it technically be possible to train HANNA directly in Clapeyron which the way you've implemented the parameters?

se-schmitt:
Hi Pierre, thanks for the fix. This might lead to wrong results in some cases (e.g. in the water-isopranol example) as HANNA needs canonical smiles to work properly. A function to canonicalize smiles would be helpful (sth. like [here](https://github.com/tspecht93/HANNA/blob/a41075440eaa3625704b99d1d7dc0f4dc96bf4e9/utils/Utils.py#L67)). Is there similar functionality in Julia?

1. The idea was to force the user to define the pure (saturation) model if required (and otherwise init just the raw activity coefficient model). A pure model like PR might fail in many cases as the database doesn't contain the required molecule. For future, a SMILES based vapor pressure model would be the best choice.
2. I'm not sure if this is possible. It would require some modifications to be compatible with the optimisers (e.g. define the trainable parameters etc.).

pw0908:
@longemen3000 Does ChemicalIdentifiers spit out the canonical smiles? Otherwise we can quickly convert our smiles to canonical using RDKit.

@se-schmitt We'll try to standardize our smiles. As for the puremodel of HANNA, the main reason I'd push for having PR being the default is just so users can immediately carry out VLE calculations. That being said, we have been discussing modifying how activity coefficient models are treated in Clapeyron as they should be useable without any other models. There are plenty of cases you'd like to use an activity coefficient model without needing the vapour phase. 

As for my other questions, makes sense. Would be interesting to revisit it in the future.

longemen3000:
on canonical SMILES, I remember that there were some problems with isomers and components that are really mixtures of isomers.

pw0908:
I managed to convert most of them in the last commit. I only ran into issues with cases where, to satisfy the valency, charges had to be added to the atoms.

longemen3000:
for example, the canonical smiles of `trans-4,4-dimethyl-2-pentene` and `cis-4,4-dimethyl-2-pentene` seems to be the same. there are a list of 38 canonical smiles that are repeated (ignoring missing ones), maybe we could just add a canonical_smiles field to identifiers.csv?

pw0908:
Ah, I see. Ok, that makes sense. I can add that as a column

se-schmitt:
Using canonical smiles to the database really simplifies this, thanks for adding them!

@pw0908 On that compilation issue: CUDA shouldnt be loaded/used by default. Might the error come from the use of JLD2.jl? I'm not sure how compatible these files are across platforms. Might make sence to change this in the future.

Should it be `ClapeyronHANNA.HANNA` in the docs ([here](https://github.com/ClapeyronThermo/Clapeyron.jl/blob/cfdaac6dd1197ab0f5077474c22f674e4d485d9e/docs/src/eos/activity.md?plain=1#L46)).

longemen3000:
Yeah, it should be `ClapeyronHANNA.HANNA`

on the registration and CI
- registration: underway (https://github.com/JuliaRegistries/General/pull/115808). the julia registrator bot already works without problem
- CI: we need to modify our current CI workflow, (or maybe the runtests.jl in the main repository?), but all that should be done in the main CI file.

PR #302, opened 2024-10-15
Implementation of polar SAFT-VR Mie

sonjaamsmith:
Addition of the Gross and Vrabec dipole (2006), quadrupole (2005), and dipole-quadrupole (2008) terms to the SAFT-VR Mie framework as implemented by Cripwell et al. (2018) and Smith et al. (2020).


longemen3000:
Nice!
- There is a bunch of files named `singledata_EOS` for UNIFAC, PR, and PCSAFT, that are unused, maybe delete those?
- can you add a test checking if the EoS gives a reasonable result?

@pw0908 any comments?

pw0908:
@longemen3000 One thing @sonjaamsmith pointed out is that we use the Hudson-McCoubrey combining rule instead of the sqrt version in SAFT-VR Mie. Do you remember why we did this? I seem to recall we had a conversation about this a long time ago but can't remember why we didn't change it.

pw0908:
Found the relevant issue; it's an outstanding mistake. We switched it for SAFT-gamma Mie but not VR Mie

sonjaamsmith:
I removed those unused files and added a test.

longemen3000 (approved):

pw0908:
Was about to commit the fix to SAFT-VR mie but it's now been added to master

PR #304, opened 2024-10-19
docs: fix typos

Mo-Gul:
and improve some formatting/style.

### DIFF
--- a/docs/src/dev.md
+++ b/docs/src/dev.md
@@ -12,7 +12,7 @@ Pages = ["dev.md"]
 
 ## Deactivating precompilation
 
-From Julia 1.9 onwards, the precompilation posibilities of Julia are vastly improved, so we opt to precompile some frequently used methods. this can be disabled by calling `precompile_clapeyron!(false)`
+From Julia 1.9 onwards, the precompilation possibilities of Julia are vastly improved, so we opt to precompile some frequently used methods. this can be disabled by calling `precompile_clapeyron!(false)`
 
 ```@docs
 Clapeyron.precompile_clapeyron!
--- a/docs/src/eos/correlations.md
+++ b/docs/src/eos/correlations.md
@@ -61,7 +61,7 @@ EoSVirial2
 
 # Solid Models
 
-Solid models provide simple approximations to the excess chemical potential in the solid phase. Intended to be used in conjuction with a liquid model within a CompositeModel
+Solid models provide simple approximations to the excess chemical potential in the solid phase. Intended to be used in conjunction with a liquid model within a CompositeModel
 
 ```@docs
 SolidHfus
--- a/docs/src/eos/correlations.md
+++ b/docs/src/eos/correlations.md
@@ -36,7 +36,7 @@ Clapeyron.DIPPR101Sat
 ```
 
 # Liquid Volume Correlations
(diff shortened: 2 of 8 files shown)


Mo-Gul:
Other things can be harmonized as well, e.g. the use of `d` in `\int ... dx`. There is one instance of an upright `d`

https://github.com/ClapeyronThermo/Clapeyron.jl/blob/529af13b80647402253f97af317d1be9e851ba09/docs/src/theory/background.md?plain=1#L29

The other (five) instances (just) use the normal math `d` (italics).

longemen3000 (approved):

longemen3000:
Thanks!
on the use of italics vs upright d, feel free to add those changes here or in another PR. I'm going to merge this tomorrow.

Mo-Gul:
What I have also noticed is that you (sometimes) write "personal names" in small letters like

https://github.com/ClapeyronThermo/Clapeyron.jl/blob/529af13b80647402253f97af317d1be9e851ba09/docs/src/user_guide/custom_methods.md?plain=1#L47

or

https://github.com/ClapeyronThermo/Clapeyron.jl/blob/529af13b80647402253f97af317d1be9e851ba09/docs/src/properties/basic.md?plain=1#L19

Is that on purpose?

longemen3000:
> Is that on purpose? 
No haha, feel free to correct those docs.

Mo-Gul:
Ok, one more thing I have noticed. Have a look at the items in https://clapeyronthermo.github.io/Clapeyron.jl/dev/user_guide/custom_methods/#Definitions

which is the output of

https://github.com/ClapeyronThermo/Clapeyron.jl/blob/529af13b80647402253f97af317d1be9e851ba09/docs/src/user_guide/custom_methods.md?plain=1#L8-L10

I am guessing that the starts of the lines should look something similar to

```md
[`volume(model, p, T, z)`](@ref)
```

right? But I don't have a clue if this gives the expected result.

longemen3000:
> I am guessing that the starts of the lines should look something similar to

i just checked the docs for Documenter.jl, and the correct syntax for those cases is actually:
```
[volume(model,p,T,z](@ref Clapeyron.volume)
```
https://documenter.juliadocs.org/stable/man/syntax/#Named-@refs

Mo-Gul:
I'd say this is something for another PR which I leave up for you, just in case this doesn't work (directly). Are you fine with that?

longemen3000:
yeah, i will test if that works later. in the meantime, i think this PR is ready to merge.

pw0908:
Mind if I take a look before merging?

pw0908:
@Mo-Gul thanks a lot for all of this! I looked over your modifications and have no issues with any of them.

If it is of interest to you, we are modifying the docs over on a separate [branch](https://github.com/ClapeyronThermo/Clapeyron.jl/tree/docs-update). Progress has been slow, but we'd be grateful for any feedback you might provide!

Mo-Gul:
Sure, but since I just forked the master branch it would be nice to first merge this PR. After that I'll refork and have a look.

Shall I start a discussion for general things about the docs or what would be a proper place for that?

pw0908:
Done!

Discussion would probably be the best place for it.

Issue #361, opened 2025-04-16
How to specify input pairs


matthew-kapp:
I have been going through the docs and I only see examples on specifying T, p input pairs. Is it possible to specify H, p as an input pair, as an example - which is most commonly used in engineering applications?

matthew-kapp:
Is there documentation somewhere on this? Specifically, where can I find which combinations are supported by PH? As an example, SingleFluid("Hydrogen") does not work for me. Unless there is an error someone can spot? Also, it outputs NaN so I can't really debug - I just guess that this model does not exist in the PH library.

```
## Imports
using Clapeyron
import Clapeyron.PH

## Clapeyron working

# Fluid parameters
fluid_model = SingleFluid("Hydrogen")
T_in = 70               # K
p_in = 350e5           # Pa

# Works: p, T --> h
h = enthalpy(fluid_model, p_in, T_in)
println(h)

# Does not work: p, h --> T
T = PH.temperature(fluid_model, p_in, h)
println(T)
```
1744.7097697458275
NaN

Sush1090:
In the example mentioned above the pressure you have mentioned (`p_in`) is beyond the critical pressure of hydrogen. Because of this the internal function `Tproperty` will return `NaN` for safety.  So in your above example if you reduce the pressure below critical pressure the solver will work. Example

```julia
using Clapeyron
import Clapeyron.PH

fluid_model = SingleFluid("Hydrogen")
T_in = 70               # K
p_in = 1e6           # Pa

h = enthalpy(fluid_model, p_in, T_in)
T = PH.temperature(fluid_model, p_in, h)
println(T)
```

But if you want to run your case, it is possible by simple root finder for example as follow:
```julia
using Clapeyron
import Clapeyron.PH

fluid_model = SingleFluid("Hydrogen")
T_in = 70               # K
p_in = 1e6           # Pa

h = enthalpy(fluid_model, p_in, T_in)

using Roots
function my_ph_temperature(model,p,h)
    f(T) = enthalpy(model,p,T) - h
    crit_T,crit_p,_ = crit_pure(model)
    @assert p >= crit_p "Pressure is less than critical pressure -  this is for supercritical pressure"
    sol = find_zero(f,(30,300)) 
    return sol
end

T_roots = my_ph_temperature(fluid_model,p_in,h)
```

Hopefully this helps.

Sush1090:
I just checked this example directly with `Tproperty`
```julia
[...]

PR #363, opened 2025-04-23
Basic docs on usage of general flash

Sush1090:
Hello,
I have added a basic doc to describe on how to use the different flash computation to compute bulk properties as mentioned in #361. 
Let me know if this needs more details or not.


longemen3000 (commented):
Seems good as a start, but maybe changing the title to explicitly mention P-H and P-S flash could help.

Also, could you mention the modules? Something along the lines of "addtionally, there are convenience modules, Clapeyron.XY, that provide calculation of properties...."

Sush1090:
Hello,
Yes I can do these changes. A couple of points:
1. I'll put example for each flash (ph,qp,vt ,...) 
2. The modules are only for available for `PS`, `PH`, and `VT`. For others we have to go through the flash manually hence I have only mentioned flash for now. Unless I am missing some implementations.

```julia
julia> using Clapeyron

julia> import Clapeyron: PS,PH,QT,QP,VT,TS
WARNING: could not import Clapeyron.QT into Main
WARNING: could not import Clapeyron.QP into Main
WARNING: could not import Clapeyron.TS into Main
```

longemen3000:
Ohhh, lol, i can add those

Sush1090:
I have two small points:

1. I think `temperature(model::EoSModel, state::FlashResult)` needs to be exported.
2. Correct me if I am wrong here, The second derivative properties have not been been implemented with flash. For example there is no implementation of `isobaric_heat_capacity(model::EoSModel, state::FlashResult)` so if one  needs to compute this he needs to get either p,T or V and then compute this manually. 

I ask this here as I might need to put this in the docs.

Sush1090:
I am sure there might be some things I have missed. Do let me know

longemen3000:
Seems good to go. I can add more info later.

longemen3000:
I just added the TS, QT, QP modules.

Also, i want your input on this issue that i got on QT and QP flashes. What do we want to do when the flash gives one phase? that should be considered a failed calculation. Do we throw an error?

Sush1090:
Can you give an example of QP/QT returning single phase? Possibly returning an error for such cases seems appropriate. 

Most of the time I have ran into NaN issues which is due to the initial point failure.

Sush1090:
Ah , I see. 
For `qt_flash` if the temperature given is beyond critical temperature then we can return error without going through any internal calculation. 
For `qp_flash` if the pressure is beyond critical pressure then we can also return error because phase change is now a singular point. 

The question becomes on how to handle edge case i.e. when q = 1. 
For example `qt_flash(model,1,Tcrit+100,z)` can return warning + property values at critical pressure or just throw generic error of out of domain.

PR #371, opened 2025-04-30
HELD Implementation

branch171:
HELD Implementation with vscode example cases. Note this current only implements tp_flash_impl in HELD_flash.jl and needs integrating into the flash code particularly tpflash and tpflash2 in PT.jl. 

tp_flash_impl returns beta,xp,vp,Gsol which can be placed in the flash returns for tpflash and tpflash2

beta,xp,vp,Gsol = Clapeyron.tp_flash_impl(model,p,T,z, HELDTPFlash(verbose = verbose))

Please test and feed back any comments or improvements


branch171:
The vscode examples are out of date and need the tp_flash_impl added I will push an update later.

src/methods/property_solvers/multicomponent/tp_flash/HELD_flash.jl
```diff
@@ -0,0 +1,1304 @@
...
+	beta,xp,vp,Gsol = HELD_impl(model,p,T,z₀,max_HELD_iters,max_trust_region_iters,tol,HELD_tol,add_near_pure_guess,add_pure_component,add_random_guess,add_all_guess,verbose)
+
+    return beta,xp,vp,Gsol
+    
+end
+
+# new HELD
+function Gershgorin(A)
+    n = size(A)[1]
+    e = Vector{Float64}(undef,n)
```
longemen3000, line 61:
```suggestion
    e = Vector{eltype(A)}(undef,n)
```

src/methods/property_solvers/multicomponent/tp_flash/HELD_flash.jl
```diff
@@ -0,0 +1,1304 @@
...
+    for j = 1:n
+	sum = 0.0
```
longemen3000, line 63:
```suggestion
	sum = zero(eltype(A))
```

src/methods/property_solvers/multicomponent/tp_flash/HELD_flash.jl
```diff
@@ -0,0 +1,1304 @@
...
+	add_all_guess,
+	verbose)
+	
+	# z₀ must sum to one, i.e. it is a mole fraction vector
+    nc = length(z₀)
+    v₀ = volume(model,p,T,z₀)
+    μ₀ = VT_chemical_potential(model,v₀,T,z₀)
+    λ₀ = (μ₀[1:nc-1] .- μ₀[nc])/R̄/T
+  # calculate reference volume based on Kays rule and vc[i] and scale to give water a vref/v ~ 1
+    pure = split_model.(model)
```
longemen3000, line 649:
```suggestion
    pure = split_pure_model(model)
```

src/methods/property_solvers/multicomponent/tp_flash/HELD_flash.jl
```diff
@@ -0,0 +1,1304 @@
...
+	end
+	if p[n] > ub[n]
+	    p[n] = ub[n]
+	end
+    return p
+end
+
+function Projection(x,lb,ub)
+    n = size(x)[1]
+    p = Vector{Float64}(undef,n)
```
longemen3000, line 255:
```suggestion
    p = Vector{Base.promote_eltype(x,lb,ub)}(undef,n)
```

longemen3000:
Hello,

Some comments.

1. i saw that a lot of the HELD code is for a box-constrained generic trust region solver, is it possible to move that code inside modules/Solvers.jl ? you can create your own file there. i would like it to wrap inside a generic solver accessible via `Solvers.optimize`

2. This is something i was talking to @pw0908 some time ago, JuMP is a quite heavy dependency to add, and while i don't feel really comfortable, i understand that this is the only way to implement HELD at the moment.

3. there are some performance optimizations missing, but nothing that should block this PR.

branch171:
Yes I think the trust region stuff could easily go into the solvers. 

But note there is some specific structures to the projections for HELD and Gibbs's that are unique to the HELD code. If those are defined in the HELD code then it will work. 

As for JuMP if there is a way to change it that would be fine. We really only need a simplex method. I used a canned code from numerical recipes c++ that worked

Any advice on optimisations appreciated.

longemen3000:
some specific optimization tips, some specific to Clapeyron, others more general:

- there is a macro (defined in Clapeyron) named `@sum`, that is equivalent to a for-loop (`@sum(x[i]*exp(y[i]))`.
- for gradients and hessians, you could use [DiffResults.jl](https://juliadiff.org/DiffResults.jl/stable/#DiffResults), or [DifferentiationInterface.jl](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/stable/api/#DifferentiationInterface.value_derivative_and_second_derivative!).  so, instead of calling `f`,`g` and `h`, one should instead call a function `fgh`, that calculates the value, gradient and hessian in one pass.
The latter one is not installed, but it is a light dependency.
- `x[1:nc]` allocates, use `@view(x[1:nc)` instead, that will create a view into the parent array.

Also, the specific return type for `tp_flash_impl` was changed, to return a `Clapeyron.FlashResult`. (this is to support other flash types). Before we had other flash types, we had the 4 argument return type for tp-flash, and for compatibility reasons, we still return those. (for a future breaking release, `tp_flash2` will be renamed to `tp_flash`).

pw0908:
It looks like you can implement HiGHS without using JuMP: 
https://ergo-code.github.io/HiGHS/dev/interfaces/julia/

Not as pretty, but more reasonable?

branch171:
I will try to use the HiGHS API and remove JuMP
And make the optimisation recommendations

I also found an annoying issue. I tested something that I thought as simple and it failed. This was air and water at 1.5 bar and 20 deg C. After a lot of frustration it turned out HELD could only solve it as water first then the air components. If the system was set up as air first then water at a certain moderately low water fraction the solver fails. It's comes up with an invalid lambda solution from the HiGHs solver that prevents the global solution. Once the components are rearranged in increasing mole fraction order it all works as expected. 

Finally the trust region solver loops is prefect for parallel solution. I'd like to implement that but it can be a future update.

pw0908:
This is something I've noticed with our other solvers as well. When you have two phases with very low concentrations, it typically prefers solving for the compositions with the order of the species in a certain order. Aside from giving guidance in the docs, I don't quite know how to handle this best?

longemen3000:
You could sort the model in the following way:
```julia
ix = sortperm(z)
model_sort = each_split_model(model, ix)
```
An equivalent way is to, instead of doing `x[nc] = 1 - sum(x)`, is doing this in the index of the component with the biggest composition, but that requires some extra care to handle the indices.

branch171:
Usingthe sorted model works but so far I have only one case where that needed to be done. Would be good to understand why its needed and if there are any other test cases that have a similar issue that we can test.

For that same problematic case if you increase the water molefraction eventually it all works with no sorting. Could it be large chemical potentials rather than small mole fractions. Needs some investigation.

longemen3000:
probably, one way check if the model sort is even necessary is to check the initial K-values, and if, for example , `!(1/n<= Kmin*Kmax <= n)`, then, perform the sorting. where n is just a constant (100?)

branch171:
If have going through some testing that the statement in the original HELD paper that the initial M set can be proven to be all that's required to initiate the method is practically not true. There are some cases where for larger component sets this does not provide a good enough cutting plane. Also in the early phase of the solution when UB and LB are far apart you need all the LB solutions below the current UB otherwise you can get very poor cutting planes that case the method to fail. Once things are tighter to the approaching hyper plane solution the stated algorithm works robustly.

branch171:
Ok what I have found is during the OPxv phase the Lambdas are very loose early on. As such for small mole fraction components you can get a sudden large Lambda that prevents the IPxv phase from finding a minimum. 

The solution is to provide an upper and lower bound on the lambdas in the OPxv phase. This works but I have no way yet of knowing how to set this at the start. Certainly (G0 -Gi)/(x0 - xi) for the initial M set is a poor estimate. It's easy once I know the answer.

longemen3000:
Hello,

I saw in some of your recent commits that you had some problems with the density solver used by GERG2008, is that right? i was just debugging #376, and there was indeed a problem with the initial value suggestions, but I'm interested if the cases are still failing (to add those cases into the test infrastructure)

branch171:
Hi

I stopped using the Clapeyron volume solution and made my own. 

The issue I had was water at 28 bath and 30 deg C. The Clapeyron solver could not identify the correct volume. 

It turns out you need to identify the spinodal points and use these as criteria to select the correct root.

The issue is the many Maxwell loops with water and selecting what looks like the minimum Gibbs and mechanically stable solution but they are inside the spinodal metastable region. 

I will post the test.

branch171:
Also I have made many improvements to HELD Code since my original PR. Not sure if you get these updates. Let me know if I need to do another PR so you can get them. 

The main one is moving the Trust Region to Clapeyron solvers and adding a trust region sub-problem solver that is more robust. It was make this update that I saw the issue with GERG2008 volumes.

branch171:
This is code that fails to solve

using Clapeyron,

components = ["carbon dioxide","nitrogen","water"]
model = GERG2008(components)

p = 64.0e5
T = 30+273.15

zs=[0.4975080785711593, 0.0049838428576813995]
z=append!(zs,1.0 - zs[1] - zs[2])

vol = volume(model,p,T,z)
mw = Clapeyron.molecular_weight(model,z)
println("mw = $(mw)")
rho = mw/vol
println("Volume = $(vol) rho = $(rho)")
rhom = mass_density(model,p,T,z)
println("rhom = $(rhom)")

branch171:
The answer should be

Volume by HELD = 3.710503849983332e-5 m3/mol
rho by HELD = 835.3971524715569 kg/m3

longemen3000:
It was absolutely related, it works with with the most recent commit. an easier way to test this is to check if the value of `Clapeyron.x0_volume(model,p,T,z,phase = :liquid)` is lower than `volume(model,p,T,z,phase = :l)`. if an initial guess for the liquid volume is outside the permissible range, empiric models tend to fail. (EoS with more theoretical bases tend to have an spinodal that allows better convergence properties)

Before:
```
julia> Clapeyron.x0_volume(model,p,T,z,phase = :l)
4.614421464979334e-5
```

After
```
julia> Clapeyron.x0_volume(model,p,T,z,phase = :l)
2.3847375780049338e-5
```

branch171:
Just as an observation for GERG2008. If you try to look at water at say 30 degrees C and 28 bar you will find the most stable volume including mechanical stability is a false root, at least from the reality of the required solution. This is because for compressed liquids the spinodal vapour root eventually disappears. For GERG2008 they probably don't care as it's a natural gas eos. But it's a pain if you want consistency across all pure components in a mixture. It means you can't dehydrate CO2 with compression and cooling. BTW false volume solutions cause a major problem to held as the initial M set is not on the true surface so the hyperplane solution comes up with a bad lambda solution. Saft eos,s seem to work well. My only issue is they are poor near the critical region and for CO2 this is important as we need to correctly predict near critical and super critical phase behaviour. I'm hoping for a cross over correction to SAFTVRMie may become available.

branch171:
Note the Clapeyron Volume solution seems to be OK. So its my solution method that is getting the wrong root.

longemen3000:
hmmm, if i remember correctly, false roots are vapour phases, right? if that is the case, and you are far from the mechanical critical point, maybe this criteria could help:
```julia
function return_true_root(model,v_liquid,v_vapour,T,x)
  gl = Clapeyron.VT0.gibbs_free_energy(model,v_liquid,T,x)
  gv = Clapeyron.VT0.gibbs_free_energy(model,v_vapour,T,x)
  if gv < gl
    B = second_virial_coefficient(model,T,x)
    if v_vapour < -2B #approximation of the spinodal, point when second order virial model diverges
      #for low temps (< 0.9Tc) this should be true. for high temps, false roots tend to disappear.
      return v_liquid
    else
      return v_vapour
    end 
  else
    return v_liquid
  end
end
```

pw0908:
For the simplex method, I found this in julia:  https://github.com/davidagold/simplex.jl/blob/master/simplex.jl

Looks abandoned and coded in a very interesting way. Could we adapt it?

branch171:
I will look into it.

For the case of volume solution the current Clapeyron method is working well. I found a bug in mine where I wasn't actually calculating dpdV correctly and I was selecting roots that were not valid solutions.

branch171:
JuMP package has now been removed and HELD uses only the C API for HiGHS

src/methods/property_solvers/multicomponent/tp_flash/HELD_flash.jl
```diff
@@ -0,0 +1,1412 @@
...
+		println("HELD  - HELD tolerence = $(HELD_tol)")
+		println("HELD  - add_pure_guess = $(add_pure_guess)")
+		println("HELD  - add_anti_pure_guess = $(add_anti_pure_guess)")
+		println("HELD  - add_pure_component = $(add_pure_component)")
+		println("HELD  - add_random_guess = $(add_random_guess)")
+		println("HELD  - add_all_guess = $(add_all_guess)")
+	end
+	beta,xp,vp,Gsol = HELD_impl(model,p,T,z₀,max_HELD_iters,max_trust_region_iters,tol,HELD_tol,add_pure_guess,add_anti_pure_guess,add_pure_component,add_random_guess,add_all_guess,verbose)
+
+    return beta,xp,vp,Gsol
```
longemen3000, line 67:
```suggestion
    return FlashResult(xp,beta,vp,FlashData(p,T,Gsol))
```

src/methods/property_solvers/multicomponent/tp_flash/HELD_flash.jl
```diff
@@ -0,0 +1,1412 @@
...
+    v = vref/rho
+	p,dpdV = p∂p∂V(model,v,T,x₀)
+    return dpdV
+end
+
+function HELD_volume1(model,p,T,x₀,vref,rho)
+	G(x) = HELD_func_rho(model,p,T,[x,1.0-x],vref,rho) + 5.474135924873183*(0.95 - x)
+    dG(x) = Solvers.derivative(G,x)
+	ddG(x) = Solvers.derivative(dG,x)
+    return G(x₀), dG(x₀), ddG(x₀)
```
longemen3000, line 170:
```suggestion
    return Solvers.f∂f∂2f(G,x₀)
```

src/methods/property_solvers/multicomponent/tp_flash/HELD_flash.jl
```diff
@@ -0,0 +1,1412 @@
...
+end
+
+function HELD_volume2(model,p,T,x₀,vref,rho)
+	G(x) = HELD_func_rho(model,p,T,x₀,vref,x)
+	dG(x) = HELD_func_p(model,p,T,x₀,vref,x)
+	ddG(x) = HELD_func_dpdV(model,T,x₀,vref,x)
+    return G(rho), dG(rho), ddG(rho)
```
longemen3000, line 177:
```suggestion
    return Solvers.f∂f∂2f(G,rho)
```

src/modules/solvers/tr_subproblem/eigenproblems.jl
```diff
@@ -0,0 +1,135 @@
...
+		mul!(y1, C, x1)
+		mul!(y2, P, x2)
+		axpy!(-one(T), y1, y2)
+
+		mul!(y1, P, x1)
+		axpy!(-dot(q, x2)/r^2, q, y1)
+	end
+	D = LinearMap{T}(A, 2*n; ismutating=true)
+
+	(λ, V, nconv, niter, nmult, _) = eigs(-D, [C 0*I; 0*I C], nev=nev, which=:LR, tol=tol)
```
longemen3000, line 59: just a question, is it possible to replace Arpack by the solver in IterativeSolvers.jl? https://iterativesolvers.julialinearalgebra.org/dev/eigenproblems/lobpcg/ . Not blocking, but interested in the answer.

src/methods/property_solvers/multicomponent/tp_flash/HELD_flash.jl
```diff
@@ -0,0 +1,1412 @@
...
+	vref = 0.0
+	for i = 1:length(x₀)
+    	Tc,pc,vc = crit[i]
+    	vref += x₀[i]*vc
+	end
+	v₀ = HELD_density(model,p,T,x₀,vref)
+	return vref/v₀
+end
+
+function HELD_density(model,p,T,x₀,vref)
```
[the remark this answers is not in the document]
branch171: I wrote this before your fix to the Clapeyron volume solver where for GERG it sometimes failed. I will test using your solver, it did work previously just for this one case it failed. My version is robust but likely slow as it tries to bracket all roots.

src/modules/solvers/tr_subproblem/eigenproblems.jl
branch171: I will look into it, the answer is likely yes. All thats needed is the eigen values.

src/methods/property_solvers/multicomponent/tp_flash/HELD_flash.jl
```diff
@@ -0,0 +1,1412 @@
...
+function HELD_volume2(model,p,T,x₀,vref,rho)
+	G(x) = HELD_func_rho(model,p,T,x₀,vref,x)
+	dG(x) = HELD_func_p(model,p,T,x₀,vref,x)
+	ddG(x) = HELD_func_dpdV(model,T,x₀,vref,x)
+    return G(rho), dG(rho), ddG(rho)
+end
+
+function HELD_volume(model,p,T,x₀)
+	pure = split_pure_model(model)
+	crit = crit_pure.(pure)
```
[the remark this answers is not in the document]
branch171: I did not realise that this was iterative for SAFT equations. All I want is a reasonable estimate for the mixture critical volume and then use this to find the upper bound on the liquid volume for the search. I will try 3*lb_volume(model,T,z)

branch171:
We only use TRS small which only uses LinearAlgebra. So I will simplify the tr_subproblem code and remove Arpack. 

I will also try to add @view where required to minimise allocations.

I need some help on the need for deepcopy. I used this but I don't know if it is required.

longemen3000:
May i ask, what does `SAFTVRMieCP` do?

branch171:
It's nothing to do with HELD. I am trying to improve predictions near the critical point for pure components and mixtures. For my work CO2 rich mixtures that operate dense phase and close to the critical point are impotent. GERG2008 is the obvious choice but it is limited in components. SAFTVRMie is a good choice but Id like better critical point predictions. I was thinking of a crossover model like Keselev. But in the first instance I'm going try to get the Tc,pc and vc to match using a simple cubic addition to ares.

pw0908:
Small detail that might be relevant: broadly speaking, it isn't possible to develop an analytical theory that can capture both the critical point and saturation curve. That's partly why SAFT-VR Mie will always struggle capturing the critical point, as well as why you need corrections like crossover theory. 

Not to say that this wont work, but you might find some diminishing returns.

branch171:
Could I ask a question:

Why in the models when you get various Helmholtz terms do you divide by sum(z) for linear mixture models and sum(z)^2 for quadratic mixtures models. 

eg
_a_2 *= -π*ρ/(T*T)/(∑z*∑z)

longemen3000:
because the original models are defined in terms of mol fractions, whereas we work with mol amounts. we want the following equality to hold:
```julia
k = rand() #any positive number
a_res(model,V,T,z) = a_res(model,k*V,T,k*z)
```

src/modules/solvers/tr_optimizer.jl
pw0908, line 1: This module is an astounding amount of work! In its current form, isn't it more suitable to be integrated as part of NLSolvers.jl? These kinds of things will probably be more useful to the general community than just users of Clapeyron.

Also, is the trust region solver in NLSolvers.jl not suitable for our applications?

pw0908 (commented):
I took some time to review this PR. Thank you @branch171 for all this amazing work! Generally speaking, I have no issue accepting the HELD-component of this PR, even with the use of HiGHS. I do wonder if the work you've done with the trust region solvers doesn't deserve to be part of a more relevant package like NLSolvers.jl or its own package? It might get under-utilised if you leave it in Clapeyron....

As for the vscode examples, they are fine as well, although I would personally prefer that they would be formatted as jupyter notebooks (which, if you're not familiar with, we can do for you). 

The only thing that's really left beyond this is the inclusion of tests. You can take a look at how we've formatted the tests for our other tp flash algorithms to get an idea.

I also see you're now doing some work with SAFT-VR Mie; this seems a bit adjacent to this PR and, since it's only started development, I think we should leave out of the master branch? Similar for the relevant examples. You can feel free to keep working on it as a separate branch! We'd be happy to help.

pw0908:
Also, aside from SAFT theories, you could consider using advanced cubics like tcPR to handle your systems? You might have to introduce some fitting parameters to get the VLE just right, but it'll do a better job modelling the critical region than the SAFT equations. It'll struggle with volumetric and calorific properties.

branch171:
Happy to have helped. For me it would be useful to get this out and used so we can see issues and improve it further. So far it's not failed to converge on any of my test problems. But as it's local optimisation based it's not guaranteed. Your 4phase problem is a good example. If you use standard initial guesses it finds 3 phases. But diagnoses it's not converged so try's to add some random starting points which ultimately find the 4th phase which then converges. 

I still have my specific volume solver. I think it would be possible to use Clapeyrons default. Also I use a simplistic density like variable,basically vc/v. This is slow for SAFT based. Using lb_volume didn't work so I will one day try to improve this. What I will do is use a Bool to select my volume or Clapeyron's for now and default to Clapeyron. I will leave the critical parameters for now as it's robust and speed isn't too bad. 

I will add the test stuff and a consolidated Jupyter notebook for the examples.

As for SAFTVRMieCP. This is just me playing with Clapeyron try to get experience doing things. I was testing parameter estimation for example. Also changing the model code. It's just a playground to get familiar with it all. 

For the trust region it's pretty general except for the very specific box constraints which need to be enforced to ensure each step is feasible. I am happy with it overall as it's proven to be vital in getting HELD to be as robust as it is.

Issue #370, opened 2025-04-29
SAFTgammaMie fails with `Real` types instead of `Float64`: Some potential adjustments to keep Parameter types generic?


Garren-H:
The current implementation of SAFTgammaMie fails when trying to contruct the model with `Real` types. This can be traced back to a few lines:

1. https://github.com/ClapeyronThermo/Clapeyron.jl/blob/f23e0dff77457205cf3710392510763fb5adbf1c/src/models/SAFT/SAFTgammaMie/SAFTgammaMie.jl#L162 
   We can change this to take into account the type of shapefactor: 
   `MixedGCSegmentParam{Base.eltype(shapefactor)}(groups,shapefactor.values,segment.values)`
    https://github.com/ClapeyronThermo/Clapeyron.jl/blob/f23e0dff77457205cf3710392510763fb5adbf1c/src/database/params/GroupParam.jl#L264
   `oneunit(T)` is incompatable with `Real`, can maybe change to `s = ones(T, length(groups.flattenedgroups))`
   https://github.com/ClapeyronThermo/Clapeyron.jl/blob/f23e0dff77457205cf3710392510763fb5adbf1c/src/database/params/GroupParam.jl#L269
   `zero(Real)` gives `Int64` as the type. This can maybe be constructed manually using
   `values = Clapeyron.PackedVectorsOfVectors.pack([zeros(T, ng) for _ in 1:nc])` # compatable with Real types

2. https://github.com/ClapeyronThermo/Clapeyron.jl/blob/f23e0dff77457205cf3710392510763fb5adbf1c/src/models/SAFT/SAFTgammaMie/SAFTgammaMie.jl#L163
   https://github.com/ClapeyronThermo/Clapeyron.jl/blob/f23e0dff77457205cf3710392510763fb5adbf1c/src/database/combiningrules.jl#L25
 https://github.com/ClapeyronThermo/Clapeyron.jl/blob/f23e0dff77457205cf3710392510763fb5adbf1c/src/database/combiningrules/outplace.jl#L23
   Here we have a manual conversion to `float`. This can maybe be removed and just keeps `P.values`, which keep type `T`.

3. https://github.com/ClapeyronThermo/Clapeyron.jl/blob/f23e0dff77457205cf3710392510763fb5adbf1c/src/models/SAFT/SAFTgammaMie/SAFTgammaMie.jl#L167-L168
[...]

PR #372, opened 2025-05-01
Generate generic SAFTgammaParam and hence model compatability with AD

Garren-H:
Implements changes in issue https://github.com/ClapeyronThermo/Clapeyron.jl/issues/370

And exposes method for `SAFTgammaMie(::GroupParam, ::Dict{String,ClapeyronParam)` from discussion https://github.com/ClapeyronThermo/Clapeyron.jl/discussions/352

### DIFF
--- a/src/database/combiningrules/outplace.jl
+++ b/src/database/combiningrules/outplace.jl
@@ -20,7 +20,7 @@ Ignores non-diagonal entries already set.
 If a Single Parameter is passed as input, it will be converted to a Pair Parameter with `pᵢᵢ = pᵢ`.
 """
 function kij_mix(f::F,P::SingleOrPair,K = nothing) where F
-    param = PairParam(P.name,P.components,float.(P.values),P.ismissingvalues,P.sourcecsvs,P.sources)
+    param = PairParam(P.name,P.components,P.values,P.ismissingvalues,P.sourcecsvs,P.sources)
     return kij_mix!(f,param,K)
 end
 
@@ -46,6 +46,6 @@ If you pass a `SingleParam` or a vector as input for `Q`, then `Qᵢⱼ` will be
 ```
 """
 function pair_mix(f::F,P::SingleOrPair,Q::SingleOrPair) where F
-    param = PairParam(P.name,P.components,float.(P.values),P.ismissingvalues,P.sourcecsvs,P.sources)
+    param = PairParam(P.name,P.components,P.values,P.ismissingvalues,P.sourcecsvs,P.sources)
     return pair_mix!(f,param,Q)
 end
--- a/src/database/params/GroupParam.jl
+++ b/src/database/params/GroupParam.jl
@@ -261,12 +261,12 @@ function Base.show(io::IO, ::MIME"text/plain", param::MixedGCSegmentParam)
     show_pairs(io,param.components,param.values,separator)
 end
 
-function MixedGCSegmentParam{T}(group::GroupParam,s = FillArrays.Fill(oneunit(T),length(groups.flattenedgroups)),segment = FillArrays.Fill(oneunit(T),length(groups.flattenedgroups))) where T <: Number
+function MixedGCSegmentParam{T}(group::GroupParam,s = ones(T, length(groups.flattenedgroups)),segment = ones(T, length(groups.flattenedgroups))) where T <: Number
     name = "mixed segment"
     components = group.components
     nc = length(components)
     ng = length(group.flattenedgroups)
-    values = PackedVectorsOfVectors.packed_fill(zero(T),FillArrays.fill(ng,nc))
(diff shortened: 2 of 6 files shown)


Garren-H:
There was a typo. I fixed this "groups"->"group" when constructing `s` and `segment` inline (but this was not the source of the failing code). 

These errors can be tracked to a typo in the second definition of `paramtype` having `m` instead of `M`
https://github.com/ClapeyronThermo/Clapeyron.jl/blob/f23e0dff77457205cf3710392510763fb5adbf1c/src/database/ClapeyronParam.jl#L29

I have fixed this as well

src/database/params/GroupParam.jl
```diff
@@ -261,12 +261,12 @@ function Base.show(io::IO, ::MIME"text/plain", param::MixedGCSegmentParam)
...
 end
 
-function MixedGCSegmentParam{T}(group::GroupParam,s = FillArrays.Fill(oneunit(T),length(groups.flattenedgroups)),segment = FillArrays.Fill(oneunit(T),length(groups.flattenedgroups))) where T <: Number
+function MixedGCSegmentParam{T}(group::GroupParam,s = ones(T, length(group.flattenedgroups)),segment = ones(T, length(group.flattenedgroups))) where T <: Number
     name = "mixed segment"
     components = group.components
     nc = length(components)
     ng = length(group.flattenedgroups)
-    values = PackedVectorsOfVectors.packed_fill(zero(T),FillArrays.fill(ng,nc))
+    values = Clapeyron.PackedVectorsOfVectors.pack([zeros(T, ng) for _ in 1:nc])
```
longemen3000, line 269: any reason for this particular change?

src/models/SAFT/SAFTgammaMie/SAFTgammaMie.jl
```diff
@@ -239,7 +258,7 @@ function SAFTVRMie(groups::GroupParam,param::SAFTgammaMieParam,sites::SiteParam
     comp_sites = gc_to_comp_sites(sites,groups)
     comp_bondvol = gc_to_comp_sites(gc_bondvol,comp_sites)
     comp_epsilon_assoc = gc_to_comp_sites(gc_epsilon_assoc,comp_sites)
-    Mw = SingleParam("Mw",components)
+    Mw = SingleParam("Mw",components, zeros(eltype(shapefactor), length(groups.components)))
```
longemen3000, line 261:
```suggestion
    Mw = get(params,"Mw") do
      SingleParam("Mw",components, zeros(eltype(shapefactor), length(groups.components)))
    end
```

src/database/combiningrules/outplace.jl
```diff
@@ -20,7 +20,7 @@ Ignores non-diagonal entries already set.
 If a Single Parameter is passed as input, it will be converted to a Pair Parameter with `pᵢᵢ = pᵢ`.
 """
 function kij_mix(f::F,P::SingleOrPair,K = nothing) where F
-    param = PairParam(P.name,P.components,float.(P.values),P.ismissingvalues,P.sourcecsvs,P.sources)
+    param = PairParam(P.name,P.components,P.values,P.ismissingvalues,P.sourcecsvs,P.sources)
```
longemen3000, line 23: This change was done because sometimes, a parameter has an integer element type, but the mixing results in a float type (resulting in errors when doing `A[i,i] = aij`)

longemen3000:
apart from the comments, can you add a simple test using the method?, i'm interested in what inputs are you using.

src/database/params/GroupParam.jl
Garren-H: `eltype(zero(Real))` -> Int 
`eltype(zeros(Real,ng))` -> Real

The previous version hence results in a packed vector of intergers if `Real` types are given. A similar thing occurs when using `oneunit`

src/models/SAFT/SAFTgammaMie/SAFTgammaMie.jl
Garren-H: This errors. Firstly, there is a typo `params` should be `param`. The other error is related to `get`:
```
get(::Clapeyron.var"#982#983"{GroupParam, SingleParam{Float64}, Vector{String}}, ::Clapeyron.SAFTgammaMieParam{Float64}, ::String)
```

Also, this shouldn't be valid right? `Mw` in params is the group molecular weights, but `Mw` is the `SAFTVRMie` function should be component molecular weights. So we only want to initialize the `SingleParam`. The component molecular weight is computed afterwards in the `SAFTgammaMie` function:
https://github.com/ClapeyronThermo/Clapeyron.jl/blob/993e5414dec369ad0d5d79b26e05682ed8045e12/src/models/SAFT/SAFTgammaMie/SAFTgammaMie.jl#L186

So there is nothing to do here?

longemen3000 (approved):

PR #386, opened 2025-06-15
Fixing typos in documents

abcdvvvv:
modified:   src/methods/property_solvers/multicomponent/flash/general_flash.jl	
modified:   src/methods/property_solvers/multicomponent/flash/QP.jl


longemen3000:
Nice!

PR #400, opened 2025-07-10
Fixed some typos in the function documentation

abcdvvvv:
I noticed that half of the documents use `[m³]`, while the other half use [`m³`]. Which one do you prefer?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "editor.formatOnSave": false
-}


longemen3000:
let's go with  `[m³]` for the moment, we can make changes faster if everything is the same, but any doc refactoring is harder if there is a mixture of different styles.

abcdvvvv:
Okay, I'll refactor them later, applying ` [m³] `

abcdvvvv:
Done.

PR #404, opened 2025-07-14
Association NRTL r_i values fix

LeoSchmider:
Changed values from r_i parameters for n-hexane and cyclohexane. Added in model code reference to paper.


longemen3000:
seems good, but, the sources for a data point should be in a `sources` column in the csv data instead of the model sources.

LeoSchmider:
The reference added to the model's code is not for the r_i data point is for the model itself the first doi is the original NRTL and the second one is from the Hao et al. article from 2020. If you want I can adapt the .csv file to the standard format you use in Clapeyron.

longemen3000:
Oh, in that case, no problem merging

PR #408, opened 2025-07-16
Docs maintain

LeoSchmider:
Mostly edits on functions descriptions, fixing typos, adds units to variables, fixes adiabatic_index units. Revision is required for file VT.jl over line 18.

### DIFF
--- a/src/methods/VT.jl
+++ b/src/methods/VT.jl
@@ -236,7 +236,7 @@ function VT_joule_thomson_coefficient(model::EoSModel, V, T, z=SA[1.])
         ∂²A∂V∂T = d²A[1,2]
         ∂²A∂V² = d²A[1,1]
         ∂²A∂T² = d²A[2,2]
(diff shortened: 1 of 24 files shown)


src/methods/VT.jl
```diff
@@ -1,20 +1,21 @@
...
-
+where A is the Helmholtz free energy `[J]`,
+V is the volume `[m³]`
 """
 function pressure(model::EoSModel, V, T, z=SA[1.])
     return -∂f∂V(model,V,T,z)
 end
 
-function temperature end
+function temperature end # This is a bit weird, I don't know if this is desired or it is just a typo.
```
longemen3000, line 18: What do you mean?

LeoSchmider: Well, the function is just empty.

longemen3000: the signature of that function is used to dispatch on other methods, for example:
https://github.com/ClapeyronThermo/Clapeyron.jl/blob/e9f9e16de05bd531caad96a931daed0d26bc3d80/src/methods/XY_methods/PH.jl#L1-L30

LeoSchmider: Ohhhh I din't know you could do that, pretty cool, thanks for the info. Gonna add new commit deleting this.

longemen3000 (approved):

Issue #419, opened 2025-07-31
Graphs in docs displays weirdly in webpage


LeoSchmider:
Hi there, 
Graphs on this [page](https://clapeyronthermo.github.io/Clapeyron.jl/dev/tutorials/binary_phase_diagrams/) display weirdly, but when I open the docs on vscode and uses the .md file preview it shows correctly. I am not sure how to fix it rather than pushing the enter key to put the text one line down. I have only seen this issue in the pT projection section but for all the graphs of the section the display problem happens.

<img width="1052" height="730" alt="Image" src="https://github.com/user-attachments/assets/384bf0cc-c773-4f1f-81ac-0073cacbc005" />

<img width="1529" height="594" alt="Image" src="https://github.com/user-attachments/assets/a17a40d5-d651-4980-ba69-101838174693" />

pw0908:
Does adding a newline fix this?

LeoSchmider:
Yes, it does. :)

PR #421, opened 2025-07-31
Doc maintain

LeoSchmider:
Doc updates, Fixes issue #419. One thing that needs to be addressed in another PR because is non trivial to solve is that depends on where you look in the package, Helmholtz free energy naming varies between f and A. I know that there is not a unic naming convention for this variable but naming should be coherent throughout the package.


longemen3000 (approved):

longemen3000:
If i remember correctly, IUPAC has some guidance about which abbreviation is correct.

PR #430, opened 2025-08-04
Add more control over specifying parameters in electrolyte eos

pw0908:
This PR should allow better custom specification of electrolyte model parameters. There are three sets of parameters that can be specified:
1. The charges of each species. @longemen3000 can users specify charges as an array in this set-up?
2. ideal_userlocations
3. neutralmodel_userlocations
4. ionmodel_userlocations
5. RSPmodel_userlocations
The only annoying part I can anticipate is that some parameters (like sigma) overlap between the ionmodel and neutralmodel. Users might need to specify them twice.

### DIFF
--- a/src/models/Electrolytes/base.jl
+++ b/src/models/Electrolytes/base.jl
@@ -37,8 +37,10 @@ function ESElectrolyte(solvents,ions;
     neutralmodel = pharmaPCSAFT,
     ionmodel = DH,
     RSPmodel = ConstRSP,
-    userlocations=String[],
-    ideal_userlocations=String[],
+    charges = String[],
+    ideal_userlocations = String[],
+    neutralmodel_userlocations = String[],
+    ionmodel_userlocations = String[],
     RSPmodel_userlocations = String[],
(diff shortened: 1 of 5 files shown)


longemen3000:
In the last revamp, we can query sigma directly from the neutral model instead of storing it in the ion model, so there is a better separation of concerns in that regard.

On the charges, do you mean a vector of integers as input?. That probably requires something like:
```
if charges isa Vector{Int}
  charges = SingleParam(charges,components,charges)
else
  params = getparams(...)
  charges = params["charges"]
```

pw0908:
In the current iteration, will it still query the neutral model to get sigma?

Also yes, that's exactly what I had in mind but I thought it would need something fancier like what we do when defining custom parameters in the REPL

longemen3000:
the current iteration uses `get_sigma(model,V,T,z)` to get the sigma parameters, and passes it to the ion model, along with the charge and RSP value.
Other ion models (like `MM1`) may overrride this.

longemen3000 (approved):

PR #440, opened 2025-08-19
Flory-Huggins

pw0908:
@LeoSchmider started the implementation here.

I still need to do some benchmarks and find some parameters to test on.

### DIFF
--- a/src/models/Activity/FH/FH.jl
+++ b/src/models/Activity/FH/FH.jl
@@ -15,7 +15,8 @@ struct FloryHuggins{c<:EoSModel} <: FloryHugginsModel
     references::Array{String,1}
 end
 
-export FloryHuggins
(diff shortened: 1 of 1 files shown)


LeoSchmider:
Hi @pw0908,
Thank you very much, you did all the heavy lifting on this 😄. 
I'm not sure I can provide you the params you are looking for. The main reference I use for activity coef is DECHEMA's book which doesn't work with FH equation.

longemen3000 (approved):
LGTM

pw0908:
Is it worth waiting for the remaining tests to pass @longemen3000 ?

longemen3000:
Each workload is executed 4 (!) times in open PRs, and the last two missing workloads are in machines that sometimes timeout.

PR #441, opened 2025-08-19
[WIP] First implementation of Nasa ideal model

LeoSchmider:
Hi there,
I've seen you wanted to implement NASA ideal model based on Glenn data. I have a code from Modelica I converted into julia as a base start. Here is what is implemented at this point. A point that needs to be discussed is that in the data provided by NASA there are ions named as components, I'm not sure how do you want to handle it.
image


longemen3000:
We already have ions as components (our electrolyte models are ion-based)

longemen3000:
some things:

- there is a 2023 paper that updates some coefficients and adds new molecules: https://doi.org/10.1016/j.jqsrt.2023.108617

- the database names needs to be standarized (use the same component names as Clapeyron, including synonyms). this is harder in this particular database due to the fact that they just have the chemical formula as identifiers. Clapeyron does have a way to standarize names (mainly for developers, as it modifies the database directly):

```julia
using Clapeyron
csv_path = "database/path/to/my/csv.csv"
data = Clapeyron.read_csv(csv_path) #reads csv using in the same way as getparams
species_old = data[:species]
species_new = Clapeyron.normalize_components_sym(species_old)
missing_species = []
for i in 1:length(species_old)
  if length(species_new[i]) != 0
    species_old[i] = species_new[i] #species_old is mutable
  else
    push!(missing_species,species_old[i])
   end
end
Clapeyron.write_csv!(csv_path ,data) #writes the db with new data back.
```
You can do this repeatedly until there is a constant amount of components left.

After that, comes the (most) tedious part: add any new components to the identifiers database. In this particular regard, `ChemicalIdentifiers.jl` could help for most of the components, but it is mostly manual input.

jjfm20:
Hi there, I would like to know if there is any progress in this specific pull, I'd like to use NASA7 or NASA9 coefficients for the properties of some solids; I could help update if told what to do; bests

longemen3000:
> Hi there, I would like to know if there is any progress in this specific pull, I'd like to use NASA7 or NASA9 coefficients for the properties of some solids; I could help update if told what to do; bests

Hello, the main blocker is updating the names to use an standardized form (`carbon dioxide` instead of `CO2`). But if you are using the model as-is, the PR should be ready

LeoSchmider:
For this model do you want the db the have both nasa and Clapeyron's naming separated via  this `~|~` or only the Clapeyron name ?

longemen3000:
> For this model do you want the db the have both nasa and Clapeyron's naming separated via this ~|~ or only the Clapeyron name ?

i would prefer both names, if its possible. If a database name is not in the identifiers database (we don't have a database of ion identifiers), just leave as is

longemen3000:
Hello, i was looking at the recent commits, and there is one particular aspect of the identifiers database that is important in the context of this PR. The identifiers are compared in normalized form, that is :(Clapeyron.normalisestring(name) == Clapeyron.normalisestring(input_name)`, so, for example `Mercury` is equivalent to `mercury`. 

That does not mean that storing `Mercury` is not allowed, but, for example, instead of:
```
Trichloroethylene~|~C2HCl3~|~C2HCL3~|~R1120~|~r1120~|~R_1120~|~r_1120~|~c2hcl3
```
This shorter identifier suffices:
```
Trichloroethylene~|~C2HCl3~|~|~R1120~|~R_1120
```

LeoSchmider:
Hi @longemen3000, tank you for the tip.
I imagine this applies for the charges db too ?
The `Ag^-´ ion doesn't have Cas nº apparently, what do you recommend doing in that case ?

longemen3000:
Hmmm, it is a good question. i would leave it blank for now. we have a few invented CAS numbers (for para-hydrogen and ortho-hydrogen), but those are mainly for CoolProp compatibility.

PR #442, opened 2025-08-27
assertions on bubble and dew methods

Sush1090:
Right now on the current master the following will work: 

```julia
julia> fluid = cPR(["propane","butane"],idealmodel = ReidIdeal);

julia> dew_pressure(fluid,300,[-1.0,-1.0])
(419532.51468221226, 9.425412399118093e-5, 0.005381927572594918, [0.23045864442659386, 0.7695413555734061])

julia> bubble_pressure(fluid,300,[-1.0,-1.0])
(616739.3091409293, 9.119630290487429e-5, 0.0035439539481169325, [0.7638732923229267, 0.23612670767707333])

julia> dew_temperature(fluid,101325,[-1.0,-1.0])
(259.48526253581366, 8.72014003143659e-5, 0.02060903027567341, [0.1697465030711994, 0.8302534969288006])

julia> bubble_temperature(fluid,101325,[-1.0,-1.0])
(243.57121051036148, 8.001888236157625e-5, 0.019359255042491082, [0.8516663757619187, 0.14833362423808133])
```

It allows the computations when moles are negative. This is not physically possible. and can return `NaN` values as there are steps which do `x/sum(x)`

```julia
julia> bubble_temperature(fluid,101325,[-1.0,1.0])
ERROR: index reduction resulted in an empty model.
```

Hence this PR puts assertions so that the input moles are positive.

### DIFF
--- a/src/methods/property_solvers/multicomponent/bubble_point.jl
+++ b/src/methods/property_solvers/multicomponent/bubble_point.jl
@@ -363,7 +363,6 @@ Returns a tuple, containing:
 By default, uses equality of chemical potentials, via [`ChemPotBubblePressure`](@ref)
 """
 function bubble_pressure(model::EoSModel,T,x;kwargs...)
-    @assert all(>(0), x) "Mole vector contains non-positive values!"
     if keys(kwargs) == (:v0,)
         nt_kwargs = NamedTuple(kwargs)
         v0 = nt_kwargs.v0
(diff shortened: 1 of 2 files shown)


longemen3000:
The errors are genuine ([0.0, 1.0] gets flagged). Also, is it possible to move the assertion into a function and put that function in base/errors.jl?

Sush1090:
I have updated it.

longemen3000 (approved):

longemen3000:
This [-1,-1] case was really particular, because `sum(z) < 0`, so `z/sum(z) > 0`.

Issue #428, opened 2025-08-04
Further performance enhancement


Sush1090:
Hello,

There seems to be a vast difference in the performance of pure vs mixtures.

```julia
using Clapeyron, BenchmarkTools

fluid = cPR(["Propane"],idealmodel = ReidIdeal); z = [1.0]
p  =101325.0
T(x) = Clapeyron.PH.temperature(fluid,p,x,z)

h = collect(range(-1,1,100));
@btime T.($h) # 1.243 ms (209 allocations: 4.30 KiB)

fluid2 = cPR(["Propane","R134a"],idealmodel = ReidIdeal); z2 = [1.0,1.0]
T2(x) =  Clapeyron.PH.temperature(fluid2,p,x,z2) # 11.435 ms (120729 allocations: 11.51 MiB)
@btime T2.($h);
nothing
```

There is almost no difference now between `Tproperty` and ` Clapeyron.PH.temperature` since 0.6.15.

I am not sure where the bottlenecks might be now.  But the difference seems significant.

abcdvvvv:
I tested the speed of v0.6.14 and 0.6.15 on my computer, Julia 1.10.10. The performance difference between pure components and mixtures has existed since 0.6.14. Furthermore, 0.6.15 has improved speed on the T(x) and T2(x) functions you provided.

For `@btime T.($h)`
0.6.14 `1.936 ms (6108 allocations: 632.39 KiB)`
0.6.15 `1.685 ms (208 allocations: 4.27 KiB)`

for `@btime T2.($h);`
0.6.14 `23.201 ms (111724 allocations: 17.22 MiB)`
0.6.15 `20.636 ms (73824 allocations: 14.50 MiB)`

The PH_property function was changed a lot.

Sush1090:
Yes indeed. That difference is due to the fixes  for #407. 
But for this I am unable to find the most significant bottlenecks as of now. I am using ProfileViewer.jl and Cthulhu.jl but generic bottlenecks are not straightforward for me to spot.

abcdvvvv:
I'm as concerned as you are about the performance of this package. I have already identified the bottleneck. I used `PProf`, and the code is as follows:

```julia
fluid2 = cPR(["Propane", "R134a"], idealmodel=ReidIdeal);
z2 = [1.0, 1.0];
T2(x) = Clapeyron.PH.temperature(fluid2, p, x, z2) # 11.435 ms (120729 allocations: 11.51 MiB)
[...]

PR #449, opened 2025-09-10
Faster `Property`/`Tproperty`

longemen3000:
This reimplements the `Tproperty`/`Tproperty` functions to calculate the edge point as the starting point.

## Algorithm
The new algorithm uses the following steps:

1. calculate edge point. if edge point is not feasible, reuse information from the failed edge point calc to obtain the mechanical critical point
2. if the property is inside the edge limits, return edge pressure/temperature
3. if the property is over the mechanical critical point, calculate the volume at specified property, compare with critical volume. calculate one bubble/dew to see if the property is between the critical volume and saturation volume. select the phase tag accordingly
4. if the edge calculation succeded and the property is at one side of the edge, calculate the corresponding bubble/dew property, select the phase tag accordingly

## Edges

This PR adds three new functions:
`edge_pressure(model,T,z)`: returns the edge pressure, liquid and vapour volumes
`edge_temperature(model,p,z)`: returns the edge temperature, liquid and vapour volumes
`mechanical_critical_point(model,z)`: mechanical critical point, where the edge vanishes.

Edge conditions are equivalent to the saturation conditions for single component models. and the mechanical critical point for single component models is equivalent to the critical point.

## performance:

For points between the edge limits, we save a lot of computation:

```julia
julia> model = cPR(["R134a","propane"],idealmodel=ReidIdeal);

julia> @time Clapeyron._Pproperty(model,320.0,0.0003,[0.5,0.5],volume)
  0.000213 seconds (279 allocations: 13.484 KiB)
(1.4084534385450277e6, :eq)
```
Even for points away from the edge, now we only need to perform 1 bubbledew, instead of two. and the initial point for bubbledew can be obtained by the initial point used for the edge calculation, so there are less allocations overall.

Based on [#428 ](https://github.com/ClapeyronThermo/Clapeyron.jl/issues/428#issuecomment-3213393315)

### DIFF
--- a/docs/src/properties/bulk.md
+++ b/docs/src/properties/bulk.md
@@ -102,6 +102,13 @@ Clapeyron.aqueous_activity
 Clapeyron.mixing
 ```
 
+## Isogibbs condition functions
+
+```@docs
+Clapeyron.edge_pressure
+Clapeyron.edge_temperature
+```
+
 ## Initial guess functions
 
 These methods are considered internal, they don't support `Symbolics.jl` or `Unitful.jl` overloads.
--- a/docs/src/properties/multi.md
+++ b/docs/src/properties/multi.md
@@ -29,6 +29,7 @@ Clapeyron.LLE
 Clapeyron.VLLE_pressure
 Clapeyron.VLLE_temperature
 Clapeyron.crit_mix
+Clapeyron.mechanical_critical_point
 Clapeyron.UCEP_mix
 Clapeyron.UCST_pressure
 Clapeyron.UCST_temperature
--- a/src/methods/initial_guess.jl
+++ b/src/methods/initial_guess.jl
@@ -978,16 +978,16 @@ end
 
 Given critical information and a temperature, extrapolate the liquid and vapor saturation volumes.
 """
-function critical_vsat_extrapolation(model,T,Tc,Vc)
+function critical_vsat_extrapolation(model,T,Tc,Vc,z = SA[1.0])
     if T > Tc
-        _0 = zero(Base.promote_eltype(model,T))
+        _0 = zero(Base.promote_eltype(model,T,z))
         nan = _0/_0
         return nan,nan
     end
     ρc = 1/Vc
     function dp(ρ,T)
-        _,dpdV = p∂p∂V(model,1/ρ,T)
-        return -dpdV*ρ*ρ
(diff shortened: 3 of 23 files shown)


longemen3000:
@Sush1090 for the supercritical region, do you have any ideas on what to do? at the moment i just calculated one bubble/dew point and just supposed equilibria, but probably a better approach could be done, knowing that the mechanical critical point location and how close it is to the mixture critical point (while being much cheaper to calculate):
```
julia> model = cPR(["R134a","propane"],idealmodel=ReidIdeal);

julia> @time mechanical_critical_point(model,[0.5,0.5])
  0.000084 seconds (55 allocations: 2.203 KiB)
(372.2030933613895, 4.1552196421954674e6, 0.00022894222018979542)

julia> @time crit_mix(model,[0.5,0.5])
  0.004637 seconds (886 allocations: 107.453 KiB)
(372.2476451898569, 4.1585724973805295e6, 0.00022893927766208215)
```

Sush1090:
Is it true that the mechanical critical point will always be less in temperature and pressure than `crit_mix`?

longemen3000:
Yes (the main phase is supercritical, but the incipient phase could still be subcritical)

Sush1090:
Yes, that was my worry. But if the difference is so small I am not sure how much difference it will be. 

Also I have another question, I was going through the new implementation:

```julia
function x0_edge_temperature(model,p,z,pure = split_pure_model(model))
  dPdTsat = extended_dpdT_temperature.(pure,p)
  T_bubble = antoine_bubble_solve(dPdTsat,p,z)
  T_dew = antoine_dew_solve(dPdTsat,p,z)
  return (T_bubble,T_dew),dPdTsat
end
```

The second output is mentioned as dew temperature, but when I do the following:
```julia
julia> model = cPR(["R134a","propane"],idealmodel=ReidIdeal);

julia> Tm,pm,vm = crit_mix(model,[0.5,0.5]);

julia> v0_edge,dpdT = Clapeyron.x0_edge_temperature(model,pm,[0.5,0.5]);

julia> v0_edge[2] - Tm
0.1013205419476435
```

Is it that the dew temperature output from `x0_edge_temperature` can be higher than critical temperature?

longemen3000:
Yes, there is a possibility for that to happen. That is why, if a convergence issue happened while calculating `edge_temperature`, then we retry using a critical point extrapolation, and try to solve the edge temperature again with the improved guesses.

PR #450, opened 2025-09-11
Python package

se-schmitt:
I added some files to make Clapeyron.jl accessible from Python through juliacall. The idea is to create a PyPi package (named `pyclapeyron` that is installable with `pip`) and ease the access to Clapeyron for Python users. 

This PR is meant to be a basis for discussion and only implements one option. It adds the following files:
- `__init__.py`: the "actual python module" - it just adds all exported functions from Clapeyron and adds them to the module namespace (otherwise, everything would be living in the jl module). This is the simplest approach without writing much Python code and thus having two documentations etc. (e.g. the "Julia" docu is also valid for the python frontend). This comes with some drawbacks related to special type conversions (e.g. input arrays need to be np arrays, output arrays need to be converted).
- `juliapkg.json`: This file just contains the julia dependencies (only Clapeyron) and allows restricting the version.
- `pyproject.toml`: Python project file

I also added some docs with an example. The Python package can be build with `python -m build` (in the Clapeyron folder) and then being installed by `pip install dist/pyclapeyron-0.1.0-py3-none-any.whl`.

```python
import pyclapeyron as cl
import numpy as np

pure = cl.SAFTVRMie("butanol")

Tc, pc, vc = cl.crit_pure(pure)
pv, vl, vv = cl.saturation_pressure(pure, 400)

rhol = cl.molar_density(pure, pv, 400, phase="liquid")
rhov = cl.molar_density(pure, pv, 400, phase="vapour")

mix = cl.PCSAFT(["propane", "ethanol"])
pv, vl, vv, _y = cl.bubble_pressure(mix, 300, np.array([0.5,0.5]))
y = np.array(_y)
```

The actual registering at PyPi could be done using GitHub actions and being triggered for every new version of Clapeyron (to make both packages have the same version). I found an example e.g. [here](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/).

However, this would be only one option - others would be:

- A "light" alternative: Just adding a section to the docs where using Clapeyron through `juliacall` is being explained. But this would be a much higher barrier to use Clapeyron from Python compared to `pip install ...`.
- A "comprehensive" alternative: Defining new API functions in the Python package that handle type conversion etc (that would be more work for implementing and maintaining).

What's your opinion on this? 

I think it would be nice to have such an interface to Python and it would make the nice models and methods available for a broader  use base.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,8 @@ LocalPreferences.toml
 /notebooks
 examples/Playground/*
 .DS_Store
+*.venv*
+*dist
+*.egg-info
+*build
+*__pycache__
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -41,6 +41,7 @@ warnonly = Documenter.except(),
                   "Macros" => "api/macros.md",
                   "Association" => "api/association.md",
                   "Parameter Estimation" => "api/estimation.md"],
+        "`pyclapeyron`" => "python.md",
         "To-do list" => "to-do_list.md"]
                   )
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -51,6 +51,12 @@ Clapeyron.jl is a registered package, it can be installed from the general regis
 pkg> add Clapeyron
 ```
 
+## Usage from Python
+
+`Clapeyron.jl` can also be used from Python via [`juliacall`](https://juliapy.github.io/PythonCall.jl/stable/juliacall/).
+Therefore, a python package named `pyclapeyron` can be installed via `pip`, i.e. `pip install pyclapeyron`.
+A detailed description of the `pyclapeyron` package is given in [Python package `pyclapeyron`](@ref).
+
 ## Citing `Clapeyron.jl`
 
 If you are using Clapeyron for your research work, please cite the following:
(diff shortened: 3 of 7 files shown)


longemen3000:
Nice!

I have some thoughts about this:
- Probably we will need to do different versioning: in addition to Clapeyron releases, there will also be API improvements in the python side.
- I would probably do the docs here, while the actual python implementation in a separate package (to allow more authors), and make the development of the package easier, because, in the long term, i would like a complete python API, but i know that takes a long time and development resources.
- for inplace functions ending in `!`,  https://github.com/SciML/diffeqpy uses the following convention:
  > To use functions which on the Julia side have a `!`, like `step!`, replace `!` by `_b` (`step!` -> `step_b`)

@pw0908  what do you think?

pw0908:
This would be awesome! Thanks @se-schmitt for initiating this! As for the practical aspects, I agree with everything Andrés said. Doing something like diffeqpy in the cleanest way to do this imo.

I also wonder, with the OOP aspect of python, could we take advantage of it and develop of a sort of cache in the python implementation? E.g. if someone calls the critical point, it gets stored.

longemen3000:
On the cache, probably we could do this on the julia side (abusing `CompositeModel` or creating a `CachedModel` wrapper)

se-schmitt:
Happy to hear that this is sth you're interested in! I'll try to further work on this based on your comments. Another repo that might be a good reference is PySR, that would be an example for a "complete" Python API. 

> for inplace functions ending in !, https://github.com/SciML/diffeqpy uses the following convention:

Yeah, it seems this is already done by juliacall and the default for inplace functions

> I also wonder, with the OOP aspect of python, could we take advantage of it and develop of a sort of cache in the python implementation? E.g. if someone calls the critical point, it gets stored.

Something like a model class that allows `model.molar_density(p, T, z)` should not be too hard to do.

Issue #465, opened 2025-10-14
tp_flash returned incorrect phase results


abcdvvvv:
### Minimal Reproducible Example
```julia
(Please copy the mixture model from #454.)
p = 1.985550610608908e6
t = 416.6628781711617
n = [55.461373286206445, 0.09264900343401582, 7.265116936961075e-9, 8.855321114218425e-14]
Tb = bubble_temperature(mix, p, n)[1] # 386.77097059716175
Td = dew_temperature(mix, p, n)[1] # 386.80418208467535
# Note that 416.6 K is a clearly vapour state.
res = Clapeyron.tp_flash2(mix, p, t, n, RRTPFlash(equilibrium=:vle, verbose=true))
```

Output:
```julia
[ Info: K0 calculated via pure VLE correlation
[ Info: K0 = [1.6772781898736717, 0.7211712633677265, 0.3312638032994018, 0.16080065809622685]
[ Info: rachford-rice limits suggests single phase result, trying to check bubble or dew conditions
[ Info: suppossing β = 1 (dew initialization)
[ Info: initial vapour fraction = 0.9999999999999998
status0 = status = Clapeyron.RREq
status=RRVapour, K=[1.0000004259373951, 0.9998466234224412, 0.9997805269252732, 0.9996889478678108]
status=RRVapou, K=[1.0000000000298153, 0.9999999821535921, 0.9999999744739644, 0.9999999638295227]
status=RRVapour, K=[1.0000000000000036, 0.9999999999979262, 0.9999999999970335, 0.9999999999957968]
status=RRLiquid, K=[0.9999999999999996, 0.9999999999999994, 0.9999999999999992, 0.9999999999999991] <- rachfordrice <- Incorrect phase flips caused by rachfordrice_β0 [1]
[ Info: 4 SS iterations done, error(lnK) = 9.245104681809835e-12
[ Info: final K values: [0.9999999999999996, 0.9999999999999994, 0.9999999999999992, 0.9999999999999991]
[ Info: final vapour fraction: 2.220446049250313e-16
[ Info: result is single-phase (does not satisfy Rachford-Rice constraints).
[ Info: procedure converged to a single liquid phase. <- caused by rachfordrice_status [2]
Flash result at T = 416.663, p = 1.98555e6 with 2 phases:
 (x = [0.998332, 0.00166773, 1.30776e-10, 1.594e-15], β = 55.554, v = 0.00132)
[...]

PR #467, opened 2025-10-16
tp_flash: residual-based boundary projection to solve #465

abcdvvvv:
# Fix: Residual-based boundary projection for `tp_flash` near bubble/dew boundary (https://github.com/ClapeyronThermo/Clapeyron.jl/issues/465#issuecomment-3405210449)

## Summary

This PR stabilizes `tp_flash` at *exact* bubble/dew conditions with trace components by projecting the final solution onto the appropriate single-phase boundary **only when warranted by the Rachford–Rice residuals**. It prevents numerically tiny but visible vapor (or liquid) fractions (e.g., β≈2e-5) from appearing in results that are physically single-phase.

Key points:

* Post-SS/DEM **residual-based projection** using existing `rr_flash_eval`:

  * If `|F(0)| ≤ K_tol` → project to **bubble**: set β=0 and return liquid only.
  * Else if `|F(1)| ≤ K_tol` → project to **dew**: set β=1 and return vapor only.
* **Gated** to avoid overhead in normal two-phase cases:

  * Activate only when the solution is **near a boundary** (`β ≤ cbrt(K_tol)` or `1-β ≤ cbrt(K_tol)`) and the current `K` straddles 1.
  * Single-ended check: evaluate only the residual on the closer boundary (0 *or* 1).

## Problem

At bubble temperature with a trace second component (≈10⁻⁹ mole fraction), the iteration converges to a mathematically valid two-phase solution with an extremely small but **non-zero** β, due to RR geometry:

$$
F'(0)=-\sum_i z_i(K_i-1)^2 \quad(\text{dominated by the trace component if } K_1\approx 1)
$$

In the reported case,
$F(0) ≈ 1×10^{-14}$ while $F′(0) ≈ −1×10^{-9}$, so a machine-precision residual at the boundary maps to a visible interior $β ≈ −F(0)/F′(0) ≈ 10^{-5}$. This is undesirable for reporting and downstream logic that expects a single-phase result.

## Minimal Reproducible Example

```julia
p = 533_336.420062676
T = 326.01032579381365
n = [79.65506153871002, 6.124388331396414e-7, 0, 0]  # 2nd comp ~10^-9 mole fraction
res = Clapeyron.tp_flash2(mix2, p, T, n, RRTPFlash(equilibrium=:vle, verbose=true))
```

Before: `final vapour fraction ≈ 1.98e-5` despite bubble point.
After: **projected to β=0** with “procedure converged to a single liquid phase.”

## What changed (implementation)

* In `Michelsentp_flash.jl`, after SS/DEM convergence:

  * Compute proximity gate `gateβ = cbrt(K_tol)`.
  * If *near boundary* and `K` *straddles 1*, evaluate **one** endpoint residual:

    * `F(0)` if `β ≤ gateβ`, otherwise `F(1)` if `1-β ≤ gateβ`.
  * If `|F(bnd)| ≤ K_tol`, set `β` to the boundary (0 or 1), update `status` accordingly, then **single** `update_rr!` with the possibly projected β.
* Combine the final `update_rr!` with the projection block to avoid duplicate updates.
* Cache `K` extrema (`Kmin_last, Kmax_last`) during the last SS iteration and reuse them in gating (no extra scans).
* **Logs** (verbose mode):

  * `post-SS RR residual F(β) = ... (projected=true/false)`
  * `boundary check(...): |F(0)|=... |F(1)|=... F'(0)≈... β_margin=... gate=...` (diagnostic; only when the gate is entered)

## Why this works

* Uses **the same RR tolerance** `K_tol` that governs β-solve to determine if a boundary solution is valid; no new magic constants.
* Projection is purely residual-driven and only after SS/DEM has converged, so it does not mask convergence issues.
* The **gate** ensures normal two-phase cases incur virtually zero extra cost.

## Performance impact

* **Negligible** in typical runs: the block is skipped unless `β` is within `cbrt(K_tol)` of a boundary *and* `K` straddles 1.
* When triggered, only a single `rr_flash_eval` at one endpoint is performed.

# Also fix: Gibbs energy accumulation logical precedence

* **Bug**: A short-circuit form like `!non_inx[i] || isinf(lnK[i]) && (gibbs += ...)` relied on operator precedence (`&&` binds tighter than `||`), so when `!non_inx[i]` was true the RHS was skipped and `gibbs` was not accumulated
* **Fix**: Make the condition explicit (or parenthesize) so accumulation occurs when **either** predicate holds; apply to both x/y contributions:

```julia
if (!non_inx[i]) || isinf(lnK[i])
    gibbs += (1-β) * x[i] * (log(x[i]) + lnϕx[i])
end
if (!non_iny[i]) || iszero(exp(lnK[i]))
    gibbs += β * y[i] * (log(y[i]) + lnϕy[i])
end
```

### DIFF
--- a/src/methods/property_solvers/multicomponent/tp_flash/Michelsentp_flash.jl
+++ b/src/methods/property_solvers/multicomponent/tp_flash/Michelsentp_flash.jl
@@ -244,7 +244,9 @@ function tp_flash_michelsen(model::EoSModel, p, T, z, method = MichelsenTPFlash(
         lnK .= log.(K)
     end
     verbose && @info "K0 = $K"
-    _1 = one(eltype(K))
+    Ktype = eltype(K)
+    _0 = zero(Ktype)
+    _1 = one(Ktype)
     # Initial guess for phase split
     status = rachfordrice_status(K,z,non_inx,non_iny,K_tol = K_tol)
     status0 = status
@@ -256,7 +258,7 @@ function tp_flash_michelsen(model::EoSModel, p, T, z, method = MichelsenTPFlash(
     =#
     
     if status == RRLiquid
-        β = zero(_1)
+        β = _0
         if maximum(K) >= 1 #liquid phase, but there is posibility to generate a vapour composition
             verbose && @info "suppossing β = 0 (bubble initialization)"
             status = RREq
@@ -272,7 +274,7 @@ function tp_flash_michelsen(model::EoSModel, p, T, z, method = MichelsenTPFlash(
     elseif status == RREq
         β = rachfordrice(K, z; non_inx=non_inx, non_iny=non_iny,K_tol = K_tol)
     else
-        β = zero(_1)/zero(_1)
+        β = _0/_0
     end
 
     verbose && @info "initial vapour fraction = $β"
@@ -300,7 +302,6 @@ function tp_flash_michelsen(model::EoSModel, p, T, z, method = MichelsenTPFlash(
         it += 1
         itacc += 1
         lnK_old .= lnK
-        verbose && @info "$it    $β  $(round(error_lnK,sigdigits=4)) $K"
 
         x,y = update_rr!(K,β,z,x,y,non_inx,non_iny)
 
@@ -332,6 +333,7 @@ function tp_flash_michelsen(model::EoSModel, p, T, z, method = MichelsenTPFlash(
         end
         K .= exp.(lnK)
         β = rachfordrice(K, z; β0=β, non_inx=non_inx, non_iny=non_iny, K_tol = K_tol)
+        verbose && @info "$it    $β  $(round(error_lnK,sigdigits=4)) $K"
 
         if isnan(β) #try to save K? basically damping
             if rachfordrice_status(K,z,non_inx,non_iny;K_tol = K_tol) != RRTrivial
@@ -345,10 +347,10 @@ function tp_flash_michelsen(model::EoSModel, p, T, z, method = MichelsenTPFlash(
         Kmin,Kmax = K_extrema(K,non_inx,non_iny)
         if status == RRLiquid && Kmin < 1
             status = RREq
-            β = eps(eltype(β))
+            β = eps(Ktype)
         elseif status == RRVapour && Kmax > 1
             status = RREq
-            β = 1 - eps(eltype(β))
+            β = 1 - eps(Ktype)
         end
         # Computing error
         # error_lnK = sum((lnK .- lnK_old).^2)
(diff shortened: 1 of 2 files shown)


longemen3000:
Test errors are expected on julia nightly.
Can you duplicate your changes on `tp_flash/electrolyte_flash.jl` ?

abcdvvvv:
Added Safeguards and Linearized-Distance Criterion

## Safeguards 

To robustly handle the “bubble-point boundary + trace second component” scenario without misclassifying `RRTrivial`, I added the following guards around the residual-based boundary projection:

* **Gate A — Only on equilibrated two-phase:**
  Apply the projection **only when** `status == RREq`, excluding `RRTrivial`.

* **Gate B — Near-boundary proximity:**
  Require closeness to a boundary via

$$
  \beta_{\text{margin}} \equiv \min(\beta,1-\beta)\ \le\ \sqrt[3]{K_{\text{tol}}},,
$$

  so typical two-phase solutions will not go into the if block.

* **Gate C — Two-phase possibility (K straddles 1):**
 `Kmin < _1 && Kmax > _1`.

* **Gate D — Sign-guided endpoint residuals**
  Evaluate endpoint RR residuals

$$
  F(0)=\sum_i z_i,(K_i-1),\qquad
  F(1)=\sum_i z_i,\frac{K_i-1}{K_i},,
$$

  and prefer:

  * **Bubble projection** $\beta\to0$ if $|F(0)|\le K_{\text{tol}}$ and $F(1)<0$.
  * **Dew projection** $\beta\to1$ if $|F(1)|\le K_{\text{tol}}$ and $F(0)>0$.
  * If **both** endpoints qualify, use the **linearized distance** rule below to pick the closer boundary in β-space (rather than arbitrarily choosing one).

## Linearized Distance

The Rachford–Rice function is

$$
F(\beta)=\sum_i \frac{z_i,(K_i-1)}{1+\beta,(K_i-1)}.
$$

Near a boundary $\beta_{\mathrm{bnd}}\in{0,1}$, take a first-order Taylor expansion:

$$
F(\beta_{\mathrm{bnd}}+\Delta\beta)\approx
F(\beta_{\mathrm{bnd}})+F'(\beta_{\mathrm{bnd}}),\Delta\beta.
$$

To reach equilibrium ($F=0$), the required β-shift is

$$
\boxed{\ |\Delta\beta| \approx \frac{|F(\beta_{\mathrm{bnd}})|}{|F'(\beta_{\mathrm{bnd}})|}\ }.
$$

We compute the endpoint slopes from

$$
F'(0) = -\sum_i z_i(K_i-1)^2,\qquad
F'(1) = -\sum_i z_i\frac{(K_i-1)^2}{K_i^2},.
$$

Define

$$
\delta\beta_0 \equiv \frac{|F(0)|}{|F'(0)|+\varepsilon},\qquad
\delta\beta_1 \equiv \frac{|F(1)|}{|F'(1)|+\varepsilon},
$$

(with $\varepsilon$ a machine-precision term in the β-type to avoid divide-by-zero).
Choose the boundary with the smaller Δβ. This selects the side that is closest in β-space, which is exactly what matters for phase fraction. It also fixes the classic pitfall of using $|F|$ alone: when the local slope $|F'|$ is extremely small (as happens with a trace component), a tiny $|F|$ can still correspond to a visible interior β.

longemen3000 (approved):

PR #470, opened 2025-10-17
MichelsenTPFlash (Stage 2): apply bound clipping to `ny` and snap `β`…

abcdvvvv:
… to phase boundaries using a small threshold.

FlashResult: update `show` to print per-phase moles as `n` (instead of `β`) to avoid confusion.

Performance enhancement: `@view(x[in_equilibria])` may create a non-contiguous (scattered) view, which degrades performance, so we precompute an index vector `equilibria_idx` for `in_equilibria` and switch to an explicit element-wise loop to maximize performance


abcdvvvv:
#465

longemen3000:
i'm gonna close this for now.

abcdvvvv:
Okay!

Issue #488, opened 2025-11-07
support for CoolProp pseudo-pure fluids


longemen3000:
from https://discourse.julialang.org/t/ad-compatible-fluid-property-backend-for-heat-pump-simulations/133658/13?u=longemen3000

PR #489, opened 2025-11-11
Pseudo-Pure CoolProp models

longemen3000:
Implementation of pseudo-pure CoolProp models.

Not much needed, apart from some parsing quirks (and storing the bubble and dew functions)

what is missing is some functions to calculate (or exclude) two-phase properties, in particular, overloading `Tproperty_pure` and `Pproperty_pure` to use the pseudo-pure bubble and dew functions instead of going for single-component saturation.


longemen3000:
fixes #488

Issue #500, opened 2025-12-01
`crit_pure` returns numeric values even when the solver does not converge


Garren-H:
Hi all I have been running a parameter estimation with `SAFTgammaMie` starting at random points. To ensure that the there exists at least 2 volume roots below the crit point (a condition to regress pure compound saturation properties), I have been implementing an initial regression using pure critical point data. The issue I have ran into is that convergence is not guaranteed (and nor should it), but numeric values are still being returned, in some instances giving negative pressures. Surely the intended behavior here should be to check for convergence and returns NaN's if not converge as is the case with other solvers such as saturation pressure?

Additionally, on this note, I have read of Michelson's paper which discusses volume solvers (https://pubs.acs.org/doi/10.1021/ie060029x, last page). Particularly, for SAFT models it may be beneficial to solve for the packing fraction instead of the volume directly. We can additionally make a parameter transformation to the unconstrained space, using the sigmoidal function, and optimize $x$ as below instead of $\eta$ directly

$$
\eta = \frac{u-l}{1+\exp(-x)}+l
$$

Here $l$ and $u$ are upper and lower bounds on $\eta$. We can set $l=0.01$ (to avoid $\eta \rightarrow 0$) and set the upper bound to either unity or 0.5 (which is more realistic). I have implemented this approach for `SAFTgammaMie`  (with $u=1.0$ and $l=0.05$) using the NLsolve.jl package (with a custom Polyalgorithm, using Trust Region first followed by the RobustMultiNewton algorithm). This approach seemed to be more stable, and does convergence where the current implementation fails (but still fails to converge in some instances)

longemen3000:
I'm interested in your implementation, is there somewhere i can see it?. I'm also interested in some failure examples to add in the tests.

On the bounded packing fraction, maybe working in terms of `Clapeyron.lb_volume` could make the solver compatible with more EoS, only SAFTs have a definition of packing fraction (cubics have covolumes, empiric multiparameter EoS have minimum volumes, etc).

If i remember correctly, `RobustMultiNewton` (from NonLinearSolve.jl), it already a mix of trust region and newton methods.

Garren-H:
> I'm interested in your implementation, is there somewhere i can see it?. I'm also interested in some failure examples to add in the tests.

[...]

PR #501, opened 2025-12-02
improve stability of `crit_pure`

longemen3000:
by working on a bounded volume transformation, mentioned in #500


longemen3000:
`BACKSAFT` is always a pain to work with, as we can't determine the "correct" fluid lower bound volume (the lower bound volume that corresponds to the close packing limit is in a false region of the PV diagram). While debugging that EoS, i managed to add multicomponent support. (the multicomponent extension is from other group, but close enough)

Apart from that, the critical solver seems to be an improvement.

src/methods/initial_guess.jl
```diff
@@ -381,8 +385,14 @@ function x0_sat_pure_near0(model, T, vl0 = volume(model,zero(T),T,phase = :l);B
     ares = a_res(model, vl0, T, z)
     lnϕ_liq0 = ares - 1 + log(RT/vl0)
     p = exp(lnϕ_liq0)
-    vv = volume_virial(B,p,T)
     pB = -0.25*RT/B
+    if lnϕ_liq0 < log(eps(eltype(T)))
+        @show "here"
```
longemen3000:
```suggestion
```

PR #503, opened 2025-12-03
Some documentation fixes

se-schmitt:
Searched for missing docstrings and (hopefully) fixed most of them.

`ActivityBubbleTemperature` and `ActivityBubbleTemperature` are missing in the code (maybe removed at some point without removing them in the docs?).


longemen3000:
yeah, they were removed recently. before, they were just worse versions of `FugBubbleTemperature` and `FugDewTemperature`. `ActivityBubblePressure` and `ActivityDewPressure` are more useful (in the sense that they are faster than `FugBubblePressure`/`FugDewPressure` for activities in low temp/approximately ideal conditions)

Issue #502, opened 2025-12-03
Correct and more efficient AD approach to remove itterative computations from an AD stack


Garren-H:
Hi all

I have looked extensively through the current AD implementations for parameter estimation, trying to figure out 
1. Why exactly the final Newton step "works"
2. Finding an alternative approach to the current method

This has lead me to the following conclusion, which should probably be documented in the docs. The final Newton update step only works for first order derivatives. Higher order derivative will always return incorrect derivatives due to the implicit assumption of Newton's method, a first order Taylor series approximation. 

# Implicit function theorem
The correct approach to computing gradients would be through the implicit function theorem. In a nut shell, the gradients of a function $\bf{F}(\bf{x}(\bf{\theta}),\bf{\theta})=\bf{0}$, where $\bf{x}$ is an implicit function of $\bf{\theta}$ is given by 

$$
0 = \frac{\partial \bf{F}}{\partial \bf{x}} \frac{d \bf{x}}{d \bf{\theta}} + \frac{\partial \bf{F}}{\partial \bf{\theta}} \rightarrow \frac{d \bf{x}}{d \bf{\theta}} = - \left(\frac{\partial \bf{F}}{\partial \bf{x}}\right)^{-1} \frac{\partial \bf{F}}{\partial \bf{\theta}}
$$

From this function, higher order derivatives can be obtained through a combination of the chain and product rules and once again the implicit function theorem (as each partial derivative above is a function of both $\bf{x}$ and $\bf{\theta}$). 

# Why Newton's method's works
We can only recover the first order derivative from Newton's method by considering the update step:

$$
\bf{x}(\bf{\theta}) = \bf{x}^\* - \left . \frac{\partial \bf{F}}{\partial \bf{x}} \right \vert_{\bf{x}=\bf{x}^\*} \bf{F}(\bf{x}^\*,\bf{\theta})
$$

Where all terms above are only function of $\bf{\theta}$ since the root $\bf{x}^\*$ is constant. The derivative here then simplifies to

$$
\frac{d \bf{x}}{d \bf{\theta}} = - \left( \frac{\partial}{\partial \bf{\theta}}\left(\left . \frac{\partial \bf{F}}{\partial \bf{x}} \right \vert_{\bf{x}=\bf{x}^\*}\right)^{-1} \bf{F}\left(\bf{x}^\*,\bf{\theta}\right) 
[...]

PR #507, opened 2025-12-08
Implicit differentiation of root finders through implicit function theorem.

Garren-H:
Implements the changes proposed in the issue https://github.com/ClapeyronThermo/Clapeyron.jl/issues/502

### DIFF
--- a/src/methods/property_solvers/multicomponent/multicomponent.jl
+++ b/src/methods/property_solvers/multicomponent/multicomponent.jl
@@ -282,10 +282,13 @@ function bubbledew_pressure_ad(model,T,z,result,_bubble)
             else
                 _x,_y = w,z
             end
-            F1 = pressure(model,vl,T,_x) - pressure(model,vv,T,_y)
+            lnfl,pl = lnf(model,vl,T,_x)
+            lnfv,pv = lnf(model,vv,T,_y)
+            
+            F1 = pl - pv
             F2 = sum(w) - 1.0 # can exclude this restriction, but would then need additional logic to parse w (excluding one component)
-            F3 = VT_chemical_potential(model,vl,T,_x) - VT_chemical_potential(model,vv,T,_y)
-            vcat(F1,F2,F3...) # can probably be efficient with preallocation and @view but requires the common Dual type between tups and x, otherwise __gradients_for_root_finders will have the incorrect Dual type
+            F3 = lnfl - lnfv
+            vcat(F1,F2,F3) # can probably be efficient with preallocation and @view but requires the common Dual type between tups and x, otherwise __gradients_for_root_finders will have the incorrect Dual type
         end
         x_dual = __gradients_for_root_finders(x,tups,f)
         vl,vv = x_dual[1:2]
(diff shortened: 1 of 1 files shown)


longemen3000:
One good optimization to reuse the primal values created during the root finder phase, mainly because `primalval(model)` is not totally free:

```julia
julia> model = SAFTgammaMie("water")
SAFTgammaMie{BasicIdeal, Float64} with 1 component:
 "water": "H2O" => 1
Group Type: SAFTgammaMie
Contains parameters: segment, shapefactor, lambda_a, lambda_r, sigma, epsilon, epsilon_assoc, bondvol, mixed_segment

julia> dual_model = Clapeyron.promote_model(ForwardDiff.Dual{Nothing,Float64,4},model)
SAFTgammaMie{BasicIdeal, ForwardDiff.Dual{Nothing, Float64, 4}} with 1 component:
 "water": "H2O" => 1
Group Type: SAFTgammaMie
Contains parameters: segment, shapefactor, lambda_a, lambda_r, sigma, epsilon, epsilon_assoc, bondvol, mixed_segment

julia> @btime Clapeyron.primalval(dual_model)
  1.280 μs (33 allocations: 2.45 KiB)
SAFTgammaMie{BasicIdeal, Float64} with 1 component:
 "water": "H2O" => 1
Group Type: SAFTgammaMie
Contains parameters: segment, shapefactor, lambda_a, lambda_r, sigma, epsilon, epsilon_assoc, bondvol, mixed_segment
```

Garren-H:
> One good optimization to reuse the primal values created during the root finder phase, mainly because `primalval(model)` is not totally free:

I was thinking of this as well. I think we can from the start of the function create the `tups` variable and another `tups_primal` and pass both to the ad logic (would then need to add another Tuple as input the existing function). A quick test availed that allocs only gets reduced by about 6. So assuming that majority of the allocs comes from storing the intermediate results as matrices, `dx_dθ`, `∂f_∂θ_dual` and `∂f_∂θ`. I do not think we can do anything about the first two variables, but might be able to somewhat remove `∂f_∂θ`. 

I am thinking about initializing `x_dual` before we do the logic. We can then store intermediate results directly in the partials field of `x_dual` and overwriting instead of having the intermediate `∂f_∂θ`. Still thinking about a practical implementation of this, but should be possible and reduce allocs significantly

longemen3000:
@Garren-H i added some minor optimizations where possible, focused mainly in using optimized versions for `f(x,theta)`.

On allocations, my worries are mainly in single-component solvers and single component methods. One way to reduce allocations would be to write an static version of `__gradients_for_root_finders`, because both `n_params` and `n_equations` are type-level information (`n_params` is a type parameter of the dual number and `n_equations` is a type parameter is the input is an `StaticArray`). I can take care of those details after merging this PR.

Garren-H:
Thanks Andrés. Some of these things still go over my head so I shall leave that to you. I will try to add some tests at some point just to make sure:
1. Errors are handled as intended
2. Add a test to get gradients wrt to parameters of model (for critical points and saturation calculations; pure and mixtures)

Such that potential future commits do not break the AD logic

longemen3000:
I think this PR is good as it is, i can make the primal and allocation optimizations in the main branch. what do you think @Garren-H ?

Garren-H:
@longemen3000 I am happy to merge. Just not sure why some tests failed this time around. Everything succeeded in the initial commit, so might need to look into that? Additionally, think we should pre-empt failed solvers, returning NaN Duals if any of the primals are NaN since `ldiv` will fail if df/dx contains NaNs

longemen3000 (approved):

Issue #526, opened 2026-01-02
VTPR translation improvement


GustavoRodriguezM:
So I was messing around with the VTPR translations. I checked the original paper 

<img width="432" height="407" alt="Image" src="https://github.com/user-attachments/assets/b7ca540a-4b53-4fe6-a7a7-f192a5838211" />

https://pubs.acs.org/doi/pdf/10.1021/ie020047o?ref=article_openPDF

This paper seems to offer a constant translation as a first option, and only then suggests using an estimation of it. It seems the number of this estimation is the value being taken as the translation parameter in every case for Clapeyron.

I think this estimation is making the current deviations for VTPR much worse than they should be. To prove this I decided to take the cs in this same paper for n-Heptane, a compound that currently is behaving very badly for me

<img width="657" height="434" alt="Image" src="https://github.com/user-attachments/assets/b705d7b6-339a-4d32-9b54-f04ac697c37f" />

Here I compare a build from PR model using a constant translation, and the default alpha parameters for Twu 1991 already in clapeyron, with the value for SAFTVRMie, an equation that is much better for n-Heptane

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/0d9e016f-23e4-43ab-9823-45d2af8faecd" />

Whereas this is the normal VTPR function, as you can see, the performance is quite a bit worse.

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/13399ba5-efc3-4530-86f0-bd56db8c757c" />

```julia code 
using Clapeyron, PyCall
plt = pyimport("matplotlib.pyplot")
np = pyimport("numpy")


Name = "n-Heptane"
c = -0.59e-6

translation = ConstantTranslation([Name];userlocations = (;v_shift = [c]))
model = PR([Name]; alpha=TwuAlpha, translation=translation)
handle = SAFTVRMie([Name])

(Tc, Pc, Vc) = crit_pure(model)

Tmax = 3 * Tc
Tmin = 0.4 * Tc
Pmin = Pc/1000
Pmax = 500 * Pc

T = np.linspace(Tmin,Tmax,500)
P = exp10.(LinRange(log10(Pmin), log10(Pmax), 500))

den_cl = np.full((500, 500), np.nan)
den_tp = np.full((500, 500), np.nan)

vp_tp = T[T.<Tc] .- T[T.<Tc]
Tsat_tp = T[T.<Tc]
for (i,t) in enumerate(T)
[...]

PR #527, opened 2026-01-03
VTPR: improve volume translation

longemen3000:
This modification is the opossite of what we did with Patel-Teja: we use experimental values first, then use the critical volume correlation.

On the volume shift parameters, i got the "experimental" saturation liquid volumes from all pure coolprop fluids, then calculated the EoS saturation liquid volumes from `PR(fluid,alpha = TwuAlpha)`. 

The database of liquid volumes at T = 0.7Tc is available as a database, if we ever want to do the whole calculation at model instantiation time instead of beforehand.

should solve #526


longemen3000:
with the fix on the correlation, this PR is still useful (in the sense that providing the experimental volume shifts is always better), but it is not completely necessary for a purely predictive model.

PR #538, opened 2026-01-28
Add Mathias-Copeman alpha model

se-schmitt:
adds Mathias-Copeman alpha model (see e.g. https://doi.org/10.1016/0378-3812(83)80084-3 -> eq. 4)


longemen3000:
Nice!

PR #539, opened 2026-01-29
Allow non-integer values for the number of groups in GC models

se-schmitt:


longemen3000:
The main problem is that, with the change, `GroupParam` is not concrete anymore, causing performance issues on all GC EoS. 
May i ask what is the final purpose?. if the purpose is accept more number types, then we can add more conversions. Now, if you are looking to _use_ other number types (fractional amount of groups, AD over groups (??) ) , then the solution is to make `GroupParam` a parametric struct on the number of groups, and changing any other struct that depends on the assumption that `eltype(n_groups) == Int`

longemen3000:
for example, the second option would be something like:
```julia
struct GroupParam{T<:Number} <: GroupParameter
    components::Array{String,1}
    groups::Array{Array{String,1},1}
    grouptype::Symbol
    n_groups::Array{Array{Int,1},1}
    n_intergroups::Vector{Matrix{Int}}
    i_groups::Array{Array{Int,1},1}
    n_groups::Array{Array{T,1},1}
    n_intergroups::Vector{Matrix{T}}
    i_groups::Array{Array{<:Number,1},1}
    flattenedgroups::Array{String,1}
    n_flattenedgroups::Array{Array{Int,1},1}
    n_flattenedgroups::Array{Array{T,1},1}
    sourcecsvs::Array{String,1}
end
```
if you have an EoS that requires a parametric group parameter, maybe creating another group param struct could help:
```julia
struct ParametricGroupParam{T<:Number} <: GroupParameter
    components::Array{String,1}
    groups::Array{Array{String,1},1}
    grouptype::Symbol
    n_groups::Array{Array{Int,1},1}
    n_intergroups::Vector{Matrix{Int}}
    i_groups::Array{Array{Int,1},1}
    n_groups::Array{Array{T,1},1}
    n_intergroups::Vector{Matrix{T}}
    i_groups::Array{Array{<:Number,1},1}
    flattenedgroups::Array{String,1}
    n_flattenedgroups::Array{Array{Int,1},1}
    n_flattenedgroups::Array{Array{T,1},1}
    sourcecsvs::Array{String,1}
end
```

se-schmitt:
Ah I suspected sth like that (I even tried making it parametric first), thanks for being careful here! I want to allow fractional amounts of groups for GC models (currently for UNIFAC).

I will change it and make `GroupParam` parametric👍

src/database/ClapeyronParam.jl
```diff
@@ -100,8 +100,6 @@ end
 Base.eltype(p::EoSParam) = Float64
 Base.eltype(p::ParametricEoSParam{T}) where T = T
 
-const PARSED_GROUP_VECTOR_TYPE = Vector{Tuple{String, Vector{Pair{String, Int64}}}}
```
longemen3000: you can also do this:
```julia
const PARSED_GROUP_VECTOR_TYPE{T} = Vector{Tuple{String, Vector{Pair{String, T}}}} where T
```

src/database/params/SiteParam.jl
```diff
@@ -122,7 +122,7 @@ function SiteParam(pairs::Dict{String,SingleParam{Int}},allcomponentsites)
     return SiteParam(components,sites,n_sites,sourcecsvs)
 end
 
-function SiteParam(input::PARSED_GROUP_VECTOR_TYPE,sourcecsvs=String[])
+function SiteParam(input::Vector{Tuple{String, Vector{Pair{String,T}}}},sourcecsvs=String[]) where {T<:Number}
```
longemen3000: SiteParam only accepts `Int64`, so a conversion is necessary (i don't know if the constructor can handle it)

longemen3000:
Some observations and suggestions:
- is it ok if we limit the change to the UNIFAC EoS for now? the last time a i added parametric EoS, a bunch of performance issues appeared. For the rest, marking `groups::GroupParam{Int64}` should maintain current behaviour. 
- The parametric change also needs to be done on `@newmodelgc`. i'm doing some changes on how the macros in master work to make this easier. 
- Is it ok if the parameter type is shared between the parametrized `GroupParam` and a ParametricParam? that would make things easier at the promotion level.

se-schmitt:
Hi Andres, I'll implement the missing stuff next week.
 
> is it ok if we limit the change to the UNIFAC EoS for now? the last time a i added parametric EoS, a bunch of performance issues appeared. For the rest, marking groups::GroupParam{Int64} should maintain current behaviour.

I compared times for UNIFAC and SAFTgMie for basic methods (and model construction) and they were pretty much the same for all cases. Anything in mind where else performance issues could appear with this change? 

> Is it ok if the parameter type is shared between the parametrized GroupParam and a ParametricParam? that would make things easier at the promotion level.

Do you mean sth like this?
```
struct SomeModel{T}
    ...
    param::SomeParam{T}
    groups::GroupParam{T}
    ...
end
```
That would mean its nearly always GroupParam{Float64}!? Or do you mean making GroupParam a subtype of a parametric type?

pw0908:
Quick bump on my part: for reactive equilibrium with group contribution methods, it would help a lot if one could specify non-integer values of groups. E.g. you could define a carboxylate with both COOH and COO- groups and just have their relative values vary.

longemen3000:
i couldn't push it directly to the branch, but you need to add the following changes.

- In `database/params/GroupParam.jl`:

```julia
function GroupParam(components,groups,grouptype,n_groups,n_intergroups,i_groups,flattenedgroups,n_flattenedgroups,sourcecsvs)
    TT = eltype(eltype(n_groups))
    return GroupParam{TT}(components,groups,grouptype,n_groups,n_intergroups,i_groups,flattenedgroups,n_flattenedgroups,sourcecsvs)
end

function Solvers.primalval(x::GroupParam{T}) where T
    n_groups2 = Solvers.primalval_eager.(x.n_groups)
    n_intergroups2 = Solvers.primalval_eager.(x.n_intergroups)
    n_flattenedgroups2 = Solvers.primalval_eager.(x.n_flattenedgroups)
    T2 = Solvers.primal_eltype(eltype(eltype(n_groups2))
    param = GroupParam{T2}(
        x.components,
        x.groups,
        x.grouptype,
        n_groups2,
        n_intergroups2,
        x.i_groups,
        x.flattenedgroups,
        n_flattenedgroups2,
        x.sourcecsvs
    )
end

Base.eltype(param::GroupParam{T}) where T = T
Base.eltype(param::Type{<:GroupParam{T}}) where T = T

function Base.convert(::Type{GroupParam{T1}},param::GroupParam{T2}) where {T1<:Number,T2<:Number}
    n_groups2 = convert.(Vector{T1},x.n_groups)
    n_intergroups2 = convert.(Matrix{T1},x.n_intergroups)
    n_flattenedgroups2 = convert.(Vector{T1},x.n_flattenedgroups)
    param = GroupParam{T1}(
        x.components,
        x.groups,
        x.grouptype,
        n_groups2,
        n_intergroups2,
        x.i_groups,
        x.flattenedgroups,
        n_flattenedgroups2,
        x.sourcecsvs
    )
end
```

in `utils/recombine.jl`:
```julia
function Base.promote_model(::Type{T},param::GroupParam) where T
    n_groups2 = [Vector{T}(xi) for xi in x.n_groups]
    n_intergroups2 = [Matrix{T}(xi) for xi in x.n_intergroups]
    n_flattenedgroups2 = [Vector{T}(xi) for xi in x.n_flattenedgroups]
    param = GroupParam{T}(
        x.components,
        x.groups,
        x.grouptype,
        n_groups2,
        n_intergroups2,
        x.i_groups,
        x.flattenedgroups,
        n_flattenedgroups2,
        x.sourcecsvs
    )
end

```
Also, the following functions need to be changed:

- In `utils/split_model.jl`, you need to modify `each_split_model(param::GroupParam,__group,Ic,Ig)` to the following:

```julia
function each_split_model(param::GroupParam{TT},__group,Ic,Ig) where TT
    grouptype = param.grouptype
    components = param.components[Ic]
    groups = param.groups[Ic]
    n_groups = param.n_groups[Ic]
    sourcecsvs = param.sourcecsvs
    len_groups = length(param.flattenedgroups)

    flattenedgroups = param.flattenedgroups[Ig]
    i_groups = [[findfirst(isequal(group), flattenedgroups)::Int for group ∈ componentgroups] for componentgroups ∈ groups]
    n_flattenedgroups = Vector{Vector{TT}}(undef,length(Ic))

    #handling for intergroups
    n_intergroups = Vector{Matrix{TT}}(undef,length(Ic))
    empty_intergroup = Matrix{TT}(undef,(0,0))
    for (k,i) in pairs(Ic)
        pii = param.n_flattenedgroups[i]
        n_flattenedgroups[k] = pii[Ig]
        pij = param.n_intergroups[i]
        if !isempty(pij)
            n_intergroups[k] = pij[Ig,Ig]
        else
            n_intergroups[k] = empty_intergroup
        end
    end

    return GroupParam{TT}(
        components,
        groups,
        grouptype,
        n_groups,
        n_intergroups,
        i_groups,
        flattenedgroups,
        n_flattenedgroups,
        sourcecsvs)
end
```

In `database/params/GroupParam.jl`, the following functions need to be changed:

- `recombine!`: needs to account for the element type of the GroupParam in the `isassigned` conditional
- `build_gc_intragroups!` (the correct way to build an empty matrix is `Matrix{T}(undef,0,0)`, same with each intergroup)

longemen3000:
> I compared times for UNIFAC and SAFTgMie for basic methods (and model construction) and they were pretty much the same for all cases. Anything in mind where else performance issues could appear with this change?

JobackIdeal, or any methods created via `@newmodelgc`. the latest master has a placeholder on how to improve this, right here:
https://github.com/ClapeyronThermo/Clapeyron.jl/blob/1a39dafde88677f10b537674e504e806d62f2ec1/src/utils/macros.jl#L271-L284

if the struct is nonparametric, then `push!(struct_fields,:(groups::Clapeyron.GroupParam{Int64}))` needs to be done instead of `push!(struct_fields,:(groups::Clapeyron.GroupParam))`. On the parametric case, the commented code should be used.

> Do you mean sth like this?
> 
> ```
> struct SomeModel{T}
>     ...
>     param::SomeParam{T}
>     groups::GroupParam{T}
>     ...
> end
> ```
> 
> That would mean its nearly always GroupParam{Float64}!? Or do you mean making GroupParam a subtype of a parametric type?

on the UNIFAC models i mean the first implication. When evaluating a model, the element type of the model is something we use a lot, and having different structs holding different, variable element types complicates things. For example, `Base.eltype(model::EoSModel)` is defined around the element type of parameters. 

Also, `Float64` can hold exactly around `2^53 - 1` values (`Base.maxintfloat(Float64) = 9.007199254740992e15`), and most of the operations used with group parameters involve multiplying the value with a float, so there is not much to lose in terms of precision in that regard. Note that for non-parametric EoS created via macro, the default is still `GroupParam{Int64}` so there is no change there.

On the second implication, the `ParametricEoSParam` is for structs containing lists of parameters so the type hierarchy of `GroupParam` is ok, but for parametric support, there is some code that needs to be changed, as the tests are just evaluating `GroupParam{Int64}`. What could be used to test is creating a `UNIFAC` model with `BigFloat` element type, and start testing some of the following:
- `Clapeyron.split_model(model)` (phase eq algorithms use this)
- `Clapeyron.recombine!(model)` (estimation uses this)
- `Clapeyron.primalval(model)`  (AD uses this)

Most of the required changes are in the comment above, but i could be missing something.

longemen3000:
there are some errors, but i see how to fix those. @se-schmitt is ok if i merge and fix those errors in master? (i want to release a new version ASAP)

se-schmitt:
Feel free to merge and fix in master👍🏼 I have some local commits, but not finished yet.

se-schmitt:
Thanks for finishing this, @longemen3000! I'll also update pyclapeyron as soon as  the new version is released (is currently using the master).

PR #572, opened 2026-05-04
General flash for activity models

se-schmitt:
This PR implements a general flash for activity models (I tested `ph_flash` and `ps_flash` as they seem to be most relevant). I tried to use as much of the existing infrastructure as possible. The implementation might be only a starting point for more sophisticated implementations (currently a loop over inner tp_flash) and it was not extensively tested.


se-schmitt:
CC @Sush1090

longemen3000:
Errors are unrelated, let me fix master first

src/methods/property_solvers/multicomponent/flash.jl
```diff
@@ -286,6 +286,27 @@ for prop in [:enthalpy,:entropy,:internal_energy,:helmholtz_free_energy]
                 res += βi*VT0.$prop(model,vi,T,xi)
                 return res
             end
+
+            function $prop(model::ActivityModel,state::FlashResult; equilibrium=:unknown)
```
longemen3000: hmm, i don't know if this is ideal, maybe we could store the index of the vapour phase inside `FlashResult` instead? (inside the `FlashData` field, with values > 1 indicating an index, 0 being no vapour phases, -1 indicating unknown vapour phases)

longemen3000 (commented):
In general, Activity models as-is are not the correct model structure for phase equilibrium. the correct structure is `PTFlashWrapper`, so, instead of defining methods for activity models, a better approach would be to define methods for `PTFlashWrapper`.

Also, because `PTFlashWrapper` is defined in the context of a fixed T, so for methods that iteratively update the temperature, one needs to call `Clapeyron.update_temperature!(wrapper,T)` in between iterations.

Finally, i would implement this as a new flash method, instead of overloading `GeneralizedXYFlash`, because it would be useful for model types.

longemen3000:
There is also an additional discrepancy (on my part):

a `GammaPhi` model, by itself, uses the pure liquid at P,T conditions as the reference state, whereas `TPFlashWrapper` uses the saturated liquid state at T conditions (with a poynting correction). while the first one is accurate, the saturated state is more stable, and more importantly, they both differ.

In this case `TPFlashWrapper` works for holding the saturated state, so if an user wants to calculate bulk properties, maybe we should just default to generating a wrapper on each property calculation, what do you think @se-schmitt ?

se-schmitt:
Thanks for reviewing, @longemen3000! Your suggestions sound all reasonable to me. I didn't dig to deep into the different wrappers and tried to use the existing xy flash infrastructure. But I agree that using GeneralizedXYFlash is only of limited help and I think using its own method for activity-like models makes things easier. 

For my understanding: Where is `PTFlashWrapper` (and its cache) currently used advantageous? Probably pT flash as the name implies :D And I'm wondering if I used activity models wrong up to now: Am I supposed (as user) to always construct a `GammaPhi` model when using activity models? I thought just `Model(... , puremodel=PR)` would do the job already.

Regarding the implementation of a xy flash for activity coefficients:
- This would require a new flash type, say `ActivityXYFlash`, holding activity specific properties (probably very similar to `GeneralizedXYFlash` but without volumes?). And where would we "hook in" the activtiy part (to be able to still use the `ph_flash`, `ps_flash`, .. functions ...)? There is currently no dispatch on `GeneralizedXYFlash` that could be used.
- With the wrapper on each property, you mean sth like `PTXFlashWrapper` (current wrapper + property field) for using during flash right? Or the direct call by a user, like `property(wrapper, p, T, x)` (when would a user use the `PTFlashWrapper` by themselve)?

Sush1090:
Hello @se-schmitt @longemen3000 

I was checking some test cases with this implementation.  I have the following bug: 
```julia
using MLThermoProperties, Clapeyron

fluid = ["butane","propane"];z = [1.0,1.0]
model = HANNA(fluid; puremodel=PR)
p = 101325.0
Td = dew_temperature(model,p,z)[1]
T1 = Td + 5 # Pure Gas Phase  =  264.26200420537083
h1 = enthalpy(model,p,T1,z)
T_computed= Clapeyron.PH.temperature(model,p,h1,z)

@show T_computed - T1
```

```julia
T_computed - T1 = -20.33121635138866

julia> Tproperty(model,p,h1,z,enthalpy)
244.03615872553203
```

I think that for this EoS `Tproperty` is failing but I dont know why.

longemen3000:
> For my understanding: Where is `PTFlashWrapper` (and its cache) currently used advantageous? Probably pT flash as the name implies :D And I'm wondering if I used activity models wrong up to now: Am I supposed (as user) to always construct a `GammaPhi` model when using activity models? I thought just `Model(... , puremodel=PR)` would do the job already.

From the user side, this is correct, and activity model suffice. But on the evaluation level, we cache everything into a `PTFlashWrapper`. this is mainly because, if we don't do that, then we evaluate saturation pressures for all pures on each function call.

> Regarding the implementation of a xy flash for activity coefficients:

> - This would require a new flash type, say ActivityXYFlash, holding activity specific properties (probably very similar to GeneralizedXYFlash but without volumes?). And where would we "hook in" the activtiy part (to be able to still use the ph_flash, ps_flash, .. functions ...)? There is currently no dispatch on GeneralizedXYFlash that could be used.
> - With the wrapper on each property, you mean sth like PTXFlashWrapper (current wrapper + property field) for using during flash right? Or the direct call by a user, like property(wrapper, p, T, x) (when would a user use the PTFlashWrapper by themselve)?

The overload for each flash method is done at the `ph_flash_impl` method:

https://github.com/ClapeyronThermo/Clapeyron.jl/blob/7b4d0b766c4d4500cdb40c0a58690ccc17a60bd3/src/methods/property_solvers/multicomponent/flash/PH.jl#L75-L80

On the implementation of `$prop(model,flashresult)`, let me check what i can do, it will probably require some thinking about the structs and moving some things around.

Sush1090:
I think I know why this is happening (why `Tproperty` and the adjoining `ph_flash` fails).

It is because for `HANNA` does not produce a discontinuity when there is phase change on the T-H diagram. It treats phase change as smooth or does not consider phase change when we compute enthalpy. Probably the error is not in `ph_flash` or `Tproperty` but rather in enthalpy computation. 

Here is the code block: 

```julia
using MLThermoProperties, Clapeyron

fluid = ["butane","propane"];z = [1.0,1.0]
model = HANNA(fluid; puremodel=PR)
p = 101325.0
Td = dew_temperature(model,p,z)[1]
Tb = bubble_temperature(model,p,z)[1]
ΔT = 5
T_range = collect(range(Tb - ΔT, Td + ΔT, 100))

h_range = zeros(length(T_range))    

for i in eachindex(T_range)
    h_range[i] = enthalpy(model,p,T_range[i],z)
end
using Plots
fig = plot(h_range,T_range,lw = 2, marker = :o,label = "HANNA", xlabel = "Enthalpy", ylabel = "Temperature")
```
HANNA_TH_diagram


Because it does not capture any discontinuity on phase change, `Tproperty` fails as one of its first step is to catch the "edge" temperature for an estimate of which side of the phase we lie on.  

So I think the bug/error is somewhere in enthalpy computation or when we use `puremodel = PR`.

Sush1090:
Incase one wants to compare it with `cPR` 

Code block:
```julia
using MLThermoProperties, Clapeyron, Plots

fluid = ["butane","propane"];z = [1.0,1.0]
model = HANNA(fluid; puremodel=PR)
p = 101325.0
Td = dew_temperature(model,p,z)[1]
Tb = bubble_temperature(model,p,z)[1]
ΔT = 5
T_range = collect(range(Tb - ΔT, Td + ΔT, 100))

h_range = zeros(length(T_range))    

for i in eachindex(T_range)
    h_range[i] = enthalpy(model,p,T_range[i],z)
end

fig = plot(h_range,T_range,lw = 2, marker = :o,label = "HANNA", xlabel = "Enthalpy", ylabel = "Temperature")


model_pr = cPR(fluid; idealmodel = ReidIdeal)
Td_pr = dew_temperature(model_pr,p,z)[1]
Tb_pr = bubble_temperature(model_pr,p,z)[1]
T_range_pr = collect(range(Tb_pr - ΔT, Td_pr + ΔT, 100))
h_range_pr = zeros(length(T_range_pr))

for i in eachindex(T_range_pr)
    h_range_pr[i] = enthalpy(model_pr,p,T_range_pr[i],z)
end


plot!(fig,h_range_pr,T_range_pr,lw = 2, marker = :o,label = "PR Model", xlabel = "Enthalpy", ylabel = "Temperature")

```

HANNA_cPR_TH_diagram


This diagrams shows that the bubble and dew temperatures computed by both are very close but the change in enthalpy between the two points is not the same.

se-schmitt:
~Hi @Sush1090, without running the code, I think this behavior is not unexpected. HANNA is not a full EOS but uses PR for the prediction of bulk properties like (PR for pure properties and HANNA for excess properties in the liquid). So in your case, one pure comp is probably in the vapour phase and one in the liquid phase (and that's why we need to handle gE models differently here). Either the phase must be provided or a flash computation must be conducted (in which the phases are handled correctly).~

se-schmitt:
> The overload for each flash method is done at the `ph_flash_impl` method:
> 
> [Clapeyron.jl/src/methods/property_solvers/multicomponent/flash/PH.jl](https://github.com/ClapeyronThermo/Clapeyron.jl/blob/7b4d0b766c4d4500cdb40c0a58690ccc17a60bd3/src/methods/property_solvers/multicomponent/flash/PH.jl#L75-L80)

@longemen3000 Why not overloading `xy_flash` and using the existing `ph_flash_impl`? This would then automatically enable e.g. `ps_flash`. The only thing missing would be `px_flash_x0` (which already works but might beb adjusted). The `PTFlashWrapper` could then be built in `xy_flash` and be used for the actual solving. Before I start revising this: Do you think the simple `Roots.Order0` algorithm suffice for now?

Sush1090:
Hi @se-schmitt, 
>  So in your case, one pure comp is probably in the vapour phase and one in the liquid phase (and that's why we need to handle gE models differently here).
Will HANNA consider one of the components to be vapour/liquid even when the temperature provided was above the dew point and below the bubble point? 

I have a 5K gap between these two points for the case I posted. 

If I considered a 50K gap below the bubble temperature and 50K above the dew temperature just to see what happens. 

I get similar behavior: 

HANNA_cPR_TH_diagram2


Is this expected as well? My understanding of this model is negligible, so I don't know if it will detect a pure liquid or gas phase even for a larger bound. Do I need to pass the phase separately? The one I am plotting overlaps if I plot with liquid phase.  As follows:
```julia
enthalpy(model,p,T,z,phase=:liquid)
```

se-schmitt:
Ahhhh... I found the main reason for the difference: You used PR with a ideal model (Reid) and for HANNA, you only used `BasicIdeal`. When building the model like this:
```
model_pr = cPR(fluid; idealmodel = ReidIdeal)
model = HANNA(fluid; puremodel=model_pr)
```
then the difference is smaller and due to using two different models. Anyway, the PR + activity model only models the liquid phase, and the gas phase will just be model as ideal mixture.

longemen3000:
I did some modifications on the master branch, to allow this PR to be done in a more concise way:
- all PT_property methods for activity models / `GammaPhi` / `FluidCorrelation`  use the unified `TPFlashWrapper` definition, in particular, the liquid phase is now an incompressible fluid that is being approximated by the saturation properties, while the vapour phase is a gas. (volumes are decoupled from the bulk property calculation and use the inner fluid model)
- `PProperty`/`TProperty` now supports activity models / `GammaPhi` / `FluidCorrelation` , via `PTFlashWrapper`. Let me know if something fails with those functions

> Do you think the simple Roots.Order0 algorithm suffice for now?
Yes, no problems with that, the same rootfinder is used in `RRQXFlash`, but i do have some observations about how the algorithm is structured, instead of doing a full rootfinder in the property objetive , and then a full tp-flash at the current (p,T,ni) conditions, there are more options that execute a lower number of property calculations:
- Full T-P equilibration, then one property objective step
- One step in T-P equilibration (rachfordrice K-value updating, for example) and one full rootfinder on the property objective (`RRQXFlash` does this)
- A nonzero number of  T-P equilibration steps, then 1 property objective step (The `Fug` bubbledew algs do this, with a default of 10 ss steps per property objective step)

se-schmitt:
HI @longemen3000, I'm still not sure what the best way is to implement this. What would be the best function to write a `PTFlashWrapper` dispatch to enable general flash for these models?  `xy_flash(...)` (in this case we'd need also need some custom function to estimate inital T (e.g. through `px_falsh_x0(model::PTFlashWrapper, ...)` -> the current version does not work for PTFlashWrapper)? I'm also a bit confused by your recent changes, e.g. what's the purpose of `fug_ss_xy_flash!` ([here](https://github.com/ClapeyronThermo/Clapeyron.jl/blob/1aa3c84b36521d8580b74eb3a6328f21c1107f22/src/methods/property_solvers/multicomponent/flash/general_flash.jl#L1015)). It's a bit hard to understand all of the different flash-related functions... I also think this branch here is kind of broken (probably will start fresh from Master).

longemen3000:
Ohhh, `fug_ss_xy_flash was just me playing with the implementation of the fugacity bubbledew successive substitution methods, trying to transform it in a xy_flash.

I actually did not check if `px_flash_x0` is working with `PTFlashWrapper` (it should, because `Tproperty` should work and `pt_flash_x0` should work), but if it doesn't, I will have to check.

If you write a flash implementation, just suppose is it already cached, then it is just a matter of caching in the correct place

se-schmitt:
Ok, this helps, thanks! I have the following: 
```julia
using Clapeyron, Statistics

model = UNIFAC(["ethanol","water"])

T = 360.
x = [0.1,0.9]
p, vl, vv, y = bubble_pressure(model, T, x)

## PT flash
x_flsh = mean.(zip(x,y))
res = tp_flash(model, p, T, x_flsh; verbose=true)

hl = enthalpy(model, p, T, x; phase=:liquid)
hv = enthalpy(model, p, T, x; phase=:gas)
hmix = (hl+hv)/2

method = Clapeyron.init_preferred_method(ph_flash, model, (;verbose=true))
cached_model = Clapeyron.__tpflash_cache_model(model,p,300.,x,method.equilibrium)
flash0 = Clapeyron.px_flash_x0(cached_model, p, 0, x, enthalpy, method)

out = ph_flash(model, p, hmix, x; verbose=true)
```
The `px_flash_x0` worked before for the same models and conditions, but fails now (returns NaN for temp). Is this sth tha twas introduced by the PTFlashWrapper changes?

longemen3000:
i just added some modifications:
- there was a missing temperature update (when passing a temperature as an initial point for bubble_temperature / dew_temperature)
- the current K-value initial guesses where not good-enough for some activity models, so i added a forcing mechanism to create a valid K value from an invalid K value.

With those modifications:

```julia
julia> Clapeyron.px_flash_x0(cached_model, p, 0, x, enthalpy, method)
[ Info: calculating temperature via Tproperty
[ Info: property at liquid edge:                -34268.31504063803
[ Info: property at vapour edge:                7559.257175284587
[ Info: temperature at edge point:              367.60832463671994
[ Info: property between the liquid and vapour edges, in the phase change region
[ Info: property in equilibria, mostly vapour with liquid, checking bubble point
[ Info: temperature at bubble point:            360.00000000423614
[ Info: property at bubble point:               -34859.56638018601
[ Info: pseudo-liquid temperature(enthalpy) in phase change region (between edge and bubble point).
[ Info: p = 100624.3470552243, T = 366.2365142399719, equilibrium status = :eq
[ Info: using PT-flash 2-phase initial point
[ Info: K0 calculated via pure VLE correlation
[ Info: K0 = [1.4379757720008408, 0.5779058097447403]
[ Info: suppossing β = 0 (bubble initialization)
[ Info: initial vapour fraction = 2.220446049250313e-16
Flash result at T = 366.237, p = 1.00624e5 with 2 phases:
 (x = [0.1, 0.9], β = 1.0, v = 2.71441e-5)
 (x = [0.216591, 0.783409], β = 2.22045e-16, v = 0.0299281)
```

PR #575, opened 2026-05-12
Fix tc translation tc twu coeffs

LeoSchmider:
Fixed refs from tcPR.jl and tcTwu.jl.
Added all iterations for M and L params and translation param c.


longemen3000:
Errors unrelated.
By the way, I got the 4th order acentric factor polynomial from https://pmc.ncbi.nlm.nih.gov/articles/PMC11984481, in my opinion, it qualifies as the latest translation correlation from the tc-PR development group, but I like to hear a second opinion

longemen3000 (approved):

LeoSchmider:
My bad, I think you are right that the quartic regression is much better than the linear one because it gives the model much more flexibility, but because the paper you linked wasn't mentioned in the bibliography and casually matches perfectly MT's values, I thought it was just a copy and paste mistake. 
By the way are you okay with the older correlations being let next to the newest ones so that user can uncomment then or do you prefer to create new independent models the soave and soave2019 ?

longemen3000:
>  By the way are you okay with the older correlations being let next to the newest ones so that user can uncomment then or do you prefer to create new independent models the soave and soave2019 ?

If it was my decision, i would use `Soave2019` as default, the main reason we still are using the original correlations is because people just expect the original correlation when calling `PR(components)`

For that reason, to improve the visibility of the alternative alpha functions available in Clapeyron, i added a summary table
 https://clapeyronthermo.github.io/Clapeyron.jl/dev/eos/cubic/#Alpha-(%CE%B1(T))-Models:-Family,-Default-Usage,-and-Compatibility-with-Cubic-EoS

longemen3000 (approved):

PR #583, opened 2026-05-30
Estimation Revamp

longemen3000:
Inspired in the estimation framework in Langmuir.jl. At first glance, there is a lot of changes, but if you squint, the core changes are the following:
- `EstimationData` now also stores the loss and the estimator function: this allows `EstimatorData` to be used as an independent object. The struct also gains support for adding data without CSVs, we can just pass a table-like object (a named tuple, for example), an estimator method, and a loss. the method that takes a list of filepaths was moved to `estimationdata_from_csvs`. Because the estimator method works on only one estimation data object, we can specialize on the inputs and outputs, storing those as a tuple instead of a vector.
- `ToEstimate` (`EstimationModel`): `ToEstimate` is now packed with a model in a new struct  named `EstimationModel`. This struct allows a model to be indexed as if it was a vector. basically `Langmuir.from_vec`, `Langmuir.to_vec`.
`Estimation`: it is now a function, the struct that holds all the data is now named `EstimationProblem`. Incredibly, there should be no breaking changes there, apart from the renaming of the struct.

The biggest change is that all the estimation is done through a minimal API stored in the `EstimationUtils` submodule. we can now define custom estimation model types and custom estimation loss types (`EstimationData` is now a subtype of `EstimationUtils.AbstractEstimationLoss`, `EstimationModel` is a subtype of `EstimationUtils.AbstractEstimationModel`).

Two new doc pages:
- Estimation basics, that documents the new API
- Custom estimation, documents how to create a custom estimation type by using the EstimationUtils package


longemen3000:
There is one change that i want to check through you @pw0908 . Before, if no index was specified, we assumed (1,1), the docs say that we assume all, but the impl assume (1,1). The implementation on this PR allows vector estimation. for example:

```
toestimate = [
  Dict(
  param = :segment
  lower = 1.0
  upper = 6.0
]
```
Allows a nc component model to fit all segment values at once. It is a strict superset of what we supported before, but is also the only breaking change. For example, in the water-ethanol example in the notebooks, the `toestimate` input uses the fact that no indices implies (1,1) to just fit water. What do you think about this? reverting to the original behaviour is just a line change.

To support "all" parameters, implies that we can support more than one parameter per dictionary, and it is true, the code now supports this.
```
toestimate = [
  Dict(
  param = :segment
  lower = [1.0,2.0]
  upper = [6.0,12.0]
  guess = [2.5,4.5]
]

```

longemen3000:
I chose to maintain the current behavior, but add a warning that in future versions it will be changed.

pw0908:
Thanks a lot for these changes @longemen3000 and sorry that I didn't find the chance to review them. I took a look at the docs and code, and everything looks good to me! If I'm not wrong, this all means we can now use gradient-based methods for optimization?

longemen3000:
> this all means we can now use gradient-based methods for optimization?

I theory, we already supported it, but I have to check if it actually works. If it does, it deserves a mention in the docs

Issue #457, opened 2025-10-08
Dew/bubble point functions has stopped working for single substances??


B-LIE:
In the recent past, the following code worked in Clapeyron:

```julia
julia> using Clapeyron
julia> mod_H2O_pr = PR(["water"])

julia> T = 80 + 273.15
julia> p = 1e5

julia> bubble_pressure(mod_H2O_pr,T,1)
(43987.666751092685, 2.2280375877369628e-5, 0.06645941977413594, 1.0)

julia> bubble_temperature(mod_H2O_pr,p,1)
(374.2402124727243, 2.2697623673985157e-5, 0.030849393761972824, 1.0)
```
With the current version of Clapeyron, I get the following behavior:

```julia
julia> bubble_pressure(mod_H2O_pr,T,1)
ERROR: MethodError: no method matching moles_positivity(::Int64)
The function `moles_positivity` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  moles_positivity(::AbstractVector{T}) where T<:Real
```

If I add a fictitious second component and set its amount to zero, I still get a result, though:
```julia
julia> mod_mix = PR(["water", "oxygen"])
PR{BasicIdeal, PRAlpha, NoTranslation, vdW1fRule} with 2 components:
 "water"
 "oxygen"
Contains parameters: a, b, Tc, Pc, Mw

julia> bubble_pressure(mod_mix,T,[1,0])
(43987.666751092685, 2.2280375877369628e-5, 0.06645941977413594, [1.0, 0.0])
```
**Q**: I assume this is a bug, right?

Sush1090:
Hello,

I think you have to pass the mole fraction always as a vector:

```julia
julia> using Clapeyron

julia>  mod_H2O_pr = PR(["water"]);

julia> T = 80 + 273.15;

julia>  p = 1e5
100000.0

julia> bubble_pressure(mod_H2O_pr,T,[1])
(43987.666751092685, 2.2280375877369628e-5, 0.06645941977413594, [1.0])
```

`mole_positivity` was in reference to #442

B-LIE:
> I think you have to pass the mole fraction always as a vector

Ah. That is a change from a previous version where a default value was given as SVector or something.

PR #603, opened 2026-07-08
fix #602

### LINKED ISSUE #602: Behavior of `bubble_pressure` changed for quasi-pure case
In v0.6.26, the `bubble_pressure` function (and maybe others) outputs y as vector of size one for quasi-pure input parameters (e.g. `x = [1,0]` for a binary system). Seems that the internal reduction is not reverted!?

MWE:
```julia
using Clapeyron
model = PR(["ethanol", "water"])
p, vl, vv, y = bubble_pressure(model, 300., [.5,.5])
@show y
p, vl, vv, y = bubble_pressure(model, 300., [1.,0.])
@show y
```
Ouptut v0.6.25
```
y = [0.5693552964625188, 0.43064470353748113]
y = [1.0, 0.0]
```
Ouptut v0.6.26
```
y = [0.5693552964625188, 0.43064470353748113]
y = [1.0]
```
Is this intended?

abcdvvvv:
The original purpose of `27f8ca85ded7b9cf1b73ea118119c6495eb2d563` was to fix a type inconsistency regarding the fourth return value (a vector) across the "fast path" and "general path" for the four functions:
- `bubble_pressure`
- `bubble_temperature`
- `dew_pressure`, and 
- `dew_temperature` 

when `length(_model_r) == 1`. The general path returns a `Vector{eltype(x)}`, whereas the fast path originally returned the input vector `x` as-is; since `x` could be a `StaticVector`, this created a mismatch with the `Vector` type returned by the general path, leading to unstable type inference by the compiler.

This commit preserves that fix for type stability and also resolves the issue of inconsistent return vector lengths identified in #602.

Regression tests added.

### DIFF
--- a/src/methods/property_solvers/multicomponent/bubble_point.jl
+++ b/src/methods/property_solvers/multicomponent/bubble_point.jl
@@ -443,8 +443,7 @@ function bubble_pressure(model::EoSModel, T, x, method::ThermodynamicMethod)
     _model_r,idx_r = index_reduction(model,x)
     if length(_model_r)==1 && !is_pseudo_pure(model)
         (P_sat,v_l,v_v) = saturation_pressure(_model_r,T)
-        y = [one(eltype(x))]
-        return (P_sat,v_l,v_v,y)
+        return (P_sat,v_l,v_v,Vector{eltype(x)}(x))
     end
     x_r = x[idx_r]
     model_r = __tpflash_cache_model(_model_r,NaN,T,x,:vle)
@@ -643,8 +642,7 @@ function bubble_temperature(model::EoSModel, p, x, method::ThermodynamicMethod)
     _model_r,idx_r = index_reduction(model,x)
     if length(_model_r) == 1 && !is_pseudo_pure(model)
         (T_sat,v_l,v_v) = saturation_temperature(_model_r,p)
-        y = [one(eltype(x))]
-        return (T_sat,v_l,v_v,y)
+        return (T_sat,v_l,v_v,Vector{eltype(x)}(x))
     end
     x_r = x[idx_r]
     model_r = __tpflash_cache_model(_model_r,p,NaN,x,:vle)
--- a/src/methods/property_solvers/multicomponent/dew_point.jl
+++ b/src/methods/property_solvers/multicomponent/dew_point.jl
@@ -141,8 +141,7 @@ function dew_pressure(model::EoSModel, T, y, method::ThermodynamicMethod)
     _model_r,idx_r = index_reduction(model,y)
     if length(_model_r) == 1 && !is_pseudo_pure(model)
         (P_sat,v_l,v_v) = saturation_pressure(_model_r,T)
-        x = [one(eltype(y))]
-        return (P_sat,v_l,v_v,x)
+        return (P_sat,v_l,v_v,Vector{eltype(y)}(y))
     end
     y_r = y[idx_r]
     model_r = __tpflash_cache_model(_model_r,NaN,T,y,:vle)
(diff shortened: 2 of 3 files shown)


abcdvvvv:
Regarding the failed test, I saw `@test dew_pressure(vdw,T,1)[1] ≈ px rtol = 1e-6` in `Clapeyron.jl/test/test_methods_api_flash.jl:685`
So, in
```julia
function dew_pressure(model::EoSModel, T, y, method::ThermodynamicMethod)
```
, is allowing `y` to be a `Float64` type a design choice?

longemen3000:
See https://github.com/ClapeyronThermo/Clapeyron.jl/issues/457

abcdvvvv:
~You would likely need to refactor this function to support scalars.~

```julia
function dew_pressure(model::EoSModel, T, y, method::ThermodynamicMethod)
    moles_positivity(y)
    y = y/sum(y)
    T = float(T)
    verbose = get_verbosity(method)
    _model_r,idx_r = index_reduction(model,y)
    
    % ⬇ Since `_model_r` does not encode the model size in its type, this condition must be evaluated dynamically.
    % ⬇ The compiler therefore cannot eliminate the general path outside this fast path. It will infer the return type
    % ⬇ of both the fast path and the general path, and then merge them.
    if length(_model_r) == 1 && !is_pseudo_pure(model)
        (P_sat,v_l,v_v) = saturation_pressure(_model_r,T)
        x = [one(eltype(y))]
        return (P_sat,v_l,v_v,x)
    end

    ...

    (P_sat, v_l, v_v, x_r) = result
    
    % ⬇ The problem is here: `index_expansion` consistently returns a vector, which means that the fourth return value on the general path is always a vector.
    x = index_expansion(x_r,idx_r)
    converged = bubbledew_check(model,P_sat,T,v_l,v_v,x,y)
    
    ...

    if converged
        % ⬇ Here, `x` is always a vector. If the fast path is allowed to return a scalar `x`, the compiler will be unable to infer a concrete type for the returned `x`,
        % ⬇ resulting in return-type instability.
        return (P_sat, v_l, v_v, x)
    else
        nan = zero(v_l)/zero(v_l)
        x = x*nan
        return (nan,nan,nan,x)
    end
end
```

We prioritize performance and type stability; therefore, we should avoid supporting too many edge cases or widely divergent types (such as supporting both scalars and vectors simultaneously).

Personally, I do not favor supporting scalars, as it would significantly increase maintenance costs. You would have to deal with numerous type-stability issues arising from scalars—the problem encountered in this PR is a prime example.

Edited:

Okay, the scalar input requirement can be met without refactoring, but the return value must consistently be a vector.
We can replace

```julia
x = [one(eltype(y))]
return (P_sat,v_l,v_v,x)
```

with


```julia
x = y isa Real? [y] : Vector{eltype(y)}(y)
return (P_sat,v_l,v_v,x)
```
I guess.

longemen3000:
I agree with you, I want to remove that number support on the mole amounts, but sadly, that is a breaking change 🫤.

What can be done is something along the lines of

```
dew_pressure(model, T, x::Number, method::ThermodynamicMethod)
return dew_pressure(model, T, [x], method) 

function dew_pressure(model, T, x::AbstractVector, method::Thermodynamic Method) 
#actual impl
```
And restrict all other `dew_pressure` overloads to only support abstract vectors instead

abcdvvvv:
I agree with your suggestion in cases where any breaking changes need to be avoided.🥹

```
dew_pressure(model, T, x::Number, method::ThermodynamicMethod)
return dew_pressure(model, T, [x], method) 

function dew_pressure(model, T, x::AbstractVector, method::Thermodynamic Method) 
#actual impl
```

This overload is good. and suitable. I will do it later.

longemen3000:
> `dew_pressure(model, T, x::Number, method::ThermodynamicMethod) return dew_pressure(model, T, [x], method) function dew_pressure(model, T, x::AbstractVector, method::Thermodynamic Method) #actual impl`

i already did in in master, so i'm gonna merge this as is.

abcdvvvv:
Thanks.

PR #605, opened 2026-07-09
Fix bubble/dew chempot type instability

abcdvvvv:
```julia
using Clapeyron
# Case A
model = PR(["ethanol", "water"])
@code_warntype bubble_pressure(model, 300., [1., 0.])

# Case B
system1 = PCSAFT(["methanol", "cyclohexane"])
@code_warntype Clapeyron.bubble_pressure(system1, 313.15, [1., 0.])
```

PR#603 only fixes part of the type instability, these functions (e.g. `bubble_pressure`) still return `NTuple{4,Any}`.
This PR fixes the stability of return value types for **Case A**.
Now they return:
```
dew_pressure_impl -> Tuple{Float64, Float64, Float64, Vector{Float64}}
dew_pressure -> Tuple{Float64, Float64, Float64, Vector{Float64}}
dew_temperature_impl -> Tuple{Float64, Float64, Float64, Vector{Float64}}
dew_temperature -> Tuple{Float64, Float64, Float64, Vector{Float64}}
```

Regression test added.

## Cause

To put it simply, since the type value of `ForwardDiff.Chunk{min(length(v0), 8)}` in the following function is dynamically calculated, the return value is inferred to be `Any`, polluting all downstream type inferences. I made a function barrier in this PR.
```julia
    r = Solvers.nlsolve(f!, v0,
        LineSearch(Newton2(v0),BoundedLineSearch(lb,ub)),
        NLSolvers.NEqOptions(method),
        ForwardDiff.Chunk{min(length(v0), 8)}()
    )
```

ForwardDiff.jl has related issues. They have discovered that dynamic calculation of Chunk will lead to type instability, which is almost impossible to solve (dynamic and specialization are always opposites). Function barriers are a reasonable solution.

## Note

**Case B** still exhibits type instability; specifically, the PC-SAFT model triggers type instability when calculating `saturation_pressure`, whereas the PR model does not. I am still investigating the issue.

### DIFF
--- a/src/methods/property_solvers/multicomponent/bubble_point.jl
+++ b/src/methods/property_solvers/multicomponent/bubble_point.jl
@@ -443,8 +443,7 @@ function bubble_pressure(model::EoSModel, T, x, method::ThermodynamicMethod)
     _model_r,idx_r = index_reduction(model,x)
     if length(_model_r)==1 && !is_pseudo_pure(model)
         (P_sat,v_l,v_v) = saturation_pressure(_model_r,T)
-        y = [one(eltype(x))]
-        return (P_sat,v_l,v_v,y)
+        return (P_sat,v_l,v_v,Vector{eltype(x)}(x))
     end
     x_r = x[idx_r]
     model_r = __tpflash_cache_model(_model_r,NaN,T,x,:vle)
@@ -643,8 +642,7 @@ function bubble_temperature(model::EoSModel, p, x, method::ThermodynamicMethod)
     _model_r,idx_r = index_reduction(model,x)
     if length(_model_r) == 1 && !is_pseudo_pure(model)
         (T_sat,v_l,v_v) = saturation_temperature(_model_r,p)
-        y = [one(eltype(x))]
-        return (T_sat,v_l,v_v,y)
+        return (T_sat,v_l,v_v,Vector{eltype(x)}(x))
     end
     x_r = x[idx_r]
     model_r = __tpflash_cache_model(_model_r,p,NaN,x,:vle)
--- a/src/methods/property_solvers/multicomponent/bubble_point.jl
+++ b/src/methods/property_solvers/multicomponent/bubble_point.jl
@@ -435,7 +435,9 @@ function bubble_pressure(model::EoSModel,T,x;kwargs...)
     return bubble_pressure(model, T, x, method)
 end
 
-function bubble_pressure(model::EoSModel, T, x, method::ThermodynamicMethod)
+bubble_pressure(model::EoSModel, T, x::Number, method::ThermodynamicMethod) = bubble_pressure(model, T, [x], method)
+
+function bubble_pressure(model::EoSModel, T, x::AbstractVector, method::ThermodynamicMethod)
     moles_positivity(x)
     x = x/sum(x)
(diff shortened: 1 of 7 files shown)


longemen3000 (approved):

abcdvvvv:
🤔

abcdvvvv:
Does this PR failed on CI ?

longemen3000:
There is a weird http 429 error, the other one seems like a genuine error that was uncovered by the PR

abcdvvvv:
I know the reason for this->https://github.com/ClapeyronThermo/Clapeyron.jl/actions/runs/28990681297/job/86029537633

```julia
@testset "bubble/dew implicit AD" begin
    admodel = cPR(["R134a","propane"]) <--- We need to put this one outside @testset "bubble/dew implicit AD"
    ...
    @test Clapeyron.Solvers.derivative(dtge,1.0) ≈ Clapeyron.derivx(dtge,1.0) rtol = 1e-5
end

@testset "bubble/dew type stability" begin # need admodel!
    @test @inferred(bubble_pressure(admodel, 300., [0.5, 0.5])) isa Tuple{Float64,Float64,Float64,Vector{Float64}}
    @test @inferred(bubble_temperature(admodel, 300., [1.,0.])) isa Tuple{Float64,Float64,Float64,Vector{Float64}}
    @test @inferred(dew_pressure(admodel, 300., [0.,1.])) isa Tuple{Float64,Float64,Float64,Vector{Float64}}
    @test @inferred(dew_temperature(admodel, 300., [0.5,0.5])) isa Tuple{Float64,Float64,Float64,Vector{Float64}}
end
```

abcdvvvv:
A quick update 

The following inference tests pass when they are run in isolation in a fresh Julia process:

```julia
admodel = cPR(["R134a", "propane"])

@test @inferred(bubble_temperature(admodel, 300, [1.0, 0.0])) isa Tuple{Float64,Float64,Float64,Vector{Float64}}
@test @inferred(dew_temperature(admodel, 300, [0.5, 0.5])) isa Tuple{Float64,Float64,Float64,Vector{Float64}}
````

However, they do not all pass when the complete `test/test_methods_api_flash.jl` file is run.

I traced the instability in the full test run to the pure-component fast path, where inference currently sees:

```julia
saturation_temperature(_model_r, p)::Tuple{Any,Any,Any}
```

I have already fixed this `saturation_temperature` path today, but haven't PR it. I will open a separate PR shortly for the `saturation_temperature` fix.

---

The reduced reproducer for the underlying inference issue is essentially:
```julia
av1 = a∂a∂V(...)
av2 = a∂a∂V(...)
result = x0_sat_pure_crit(...)
```

**Each part is inferred concretely in isolation, but when the two consecutive `a∂a∂V` calls and `x0_sat_pure_crit` appear in the same caller, inference may widen the result of `x0_sat_pure_crit` to `Tuple{Any,Any,Any}`.**

This suggests some order- or context-dependent inference behavior, possibly related to inference caching or invalidation, incomplete recursive inference, or compiler complexity limits. I have not established the exact mechanism.

At the package level, I can mitigate the issue with targeted type assertions and small refactorings, but identifying the compiler-level root cause is considerably more difficult.

PR #606, opened 2026-07-09
Replace lnϕ with modified_lnϕ also in fallback

se-schmitt:
This should be modified_lnϕ right? If the volumes are NaN, the old version throw an error for activity models (instead of NaN for failed solve)


longemen3000:
Indeed!

PR #610, opened 2026-07-23
Fix group name inconsistency for amides and sulfones

spaghettikonfetti:
I have noticed a group name mismatch between the like and unlike parameter files for the amide and sulfone groups.
Corrected group names are now also consistent to other UNIFAC versions in Clapeyron.

Amides (changed in unlike file)
AM... --> CON...

Sulfones (changed in like file)
CH2SuCH2 --> (CH2)2SU
CH2SuCH --> CH2CHSU


longemen3000 (approved):

longemen3000:
Thanks for the fix!

PR #619, opened 2026-08-24
Format .jl files

longemen3000:
First time i've run a full formatter on the database, so i would like some feedback about the styles.
The formatter chosen was `JuliaFormatter.jl`, with the configuration stored in https://github.com/ClapeyronThermo/Clapeyron.jl/blob/master/.JuliaFormatter.toml:

I intend to use it manually (manual workflow dispatch) instead of failing pull requests for style choices.
Probably the same can be done to for spellchecking.

cc @pw0908


abcdvvvv:
You've finally decided to use it.
I noticed that many function definitions starting with `function` have been changed to the `=` definition style; is this intended?

longemen3000:
It is a specific setting (`long_to_short_function_def`)

abcdvvvv:
Oh, okay.

abcdvvvv:
I took a look at this configuration, and I have three small suggestions, just for reference.

First, you might consider turning off long_to_short_function_def. Personally, I tend to prefer preserving the function definition style chosen by the author, rather than having the formatter convert between the two forms automatically.

Second, it might be useful to set margin to a fairly large value, perhaps something like 500. Modern IDEs already provide automatic line wrapping for display, and authors can still add line breaks manually where they feel they improve readability. In some cases, automatic wrapping by the formatter can also change the intended layout of the code.

Finally, turning off whitespace_in_kwargs could make the formatting a little more compact.

longemen3000:
fair suggestions.

JuliaFormatter.jl uses for the minimal style a margin of 10000, if that is an indication.

Issue #621, opened 2026-08-25
Missing Unitful support on mass_enthalpy


waterlubber:
Good afternoon, 

I'm trying to use Clayperon for some thermal simulations after experiencing severe performance problems with the CoolProp.jl wrapper (around 10-100ms per call). 

Looks like mass_enthalpy has type errors:
```

julia> eos = Clapeyron.SingleFluid("helium")
MultiParameter Equation of state for helium:
 Polynomial power terms: 6
 Exponential terms: 6
 Gaussian bell-shaped terms: 11

julia> using Unitful: @u_str

julia> mass_enthalpy(eos, 10u"psi", 50u"K")
ERROR: MethodError: no method matching fgradf2(::Clapeyron.StaticForwardDiffTags.WithContext{…}, ::Quantity{…}, ::Quantity{…})
The function `fgradf2` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  fgradf2(::F, ::Tuple{T, T}, ::Any) where {F, T}
   @ Clapeyron C:\Users\-\.julia\packages\Clapeyron\wjRVE\src\modules\solvers\ad.jl:103
  fgradf2(::F, ::R, ::R) where {F, R<:Real}
   @ Clapeyron C:\Users\-\.julia\packages\Clapeyron\wjRVE\src\modules\solvers\ad.jl:92
  fgradf2(::F, ::R1, ::R2) where {F, R1<:Real, R2<:Real}
   @ Clapeyron C:\Users\-\.julia\packages\Clapeyron\wjRVE\src\modules\solvers\ad.jl:87
  ...

Stacktrace:
 [1] ∂f(model::SingleFluid{…}, V::Quantity{…}, T::Quantity{…}, z::StaticArraysCore.SVector{…})
   @ Clapeyron C:\Users\-\.julia\packages\Clapeyron\wjRVE\src\methods\differentials.jl:54
 [2] ∂f_vec(model::SingleFluid{…}, V::Quantity{…}, T::Quantity{…}, z::StaticArraysCore.SVector{…})
   @ Clapeyron C:\Users\-\.julia\packages\Clapeyron\wjRVE\src\methods\differentials.jl:59
 [3] VT_enthalpy(model::SingleFluid{…}, V::Quantity{…}, T::Quantity{…}, z::StaticArraysCore.SVector{…})
[...]

PR #624, opened 2026-09-01
Extended Units support (part 1)

longemen3000:
This PR changes the way how units are evaluated in Clapeyron.jl.

We now have an internal "unit" system that determines if we are using units or not.
The unit system is composed of the following functions:
- `ustrip(x,f::F) where F`  general unit stripping. for pressure, temperature, or any other property where we know exactly the units of the input. there are modified functions `uvstrip` (volume),`uzstrip` (compositions),`uhstrip` (enthalpy, internal energy) and `usstrip` (entropy) that handle inputs when more than one unit type could be used. for example, in P-H flash, total enthalpy, molar enthalpy and mass enthalpy are valid inputs, densities are a valid type of volume input and mass compositions are valid compositions.
- `unit_system(x)`: determines what unit system or package are we using. defaults to `nothing`
- `unit_type(::System,f::F)`: the expected output type of a function
- `with_output_unit(result,(::System,user_output),f::F) where F` the function that handles what is the type and units of the result. when the unit system is equal to `nothing`, then the result is returned with no changes. Functions that don't have units also skip this function.

all related unit functions are in `methods/XY_methods/units.jl`

The `Unitful.jl` extension defines their own unit system (`UnitfulJL`) and defines their unit types.

so, the `xstrip` family of functions handles the unit inputs, and `with_output_unit` handles the output units.

With that, unitful support was added to:
- all missing bulk functions (#621)
- all `XY` modules
- all flashes: `FlashResult` result types are not affected, but units can be used as input.


longemen3000:
Adding support for other unit packages (`DynamicQuantities`) is trivial, it is just a matter of copying and adapting the Unitful definitions.

What is missing from this PR:
- support for `saturation_pressure`/`saturation_temperature`: i need to do some changes to `FluidCorrelation` first

PR #626, opened 2026-09-02
feat: support mass fractions as composition input

### LINKED ISSUE #625: mass fractions as input.
with the new unitful system in place, one could tentatively add a lightweight unit system that only handles our input units. One problem that i have with Unitful.jl is that it does not handle mass fractions; a mass fraction is in the "nullspace" of units (mass/mass).  We can simulate molar fractions by passing molar amounts that sum to one, and we can pass mass amounts via Units, but mass fractions are still not handled.

If someone is interested, it just needs to overload `uzstrip` (#624)

w3lld1:
## Summary

- I added a `MassFractions` composition wrapper that converts a one-kilogram mass-fraction basis to molar amounts using each component's molecular weight.
- I covered both the conversion itself and an end-to-end property calculation alongside equivalent Unitful mass inputs.
- I included the focused test in both the standard and extension test entry points.

## Validation

- `julia --compiled-modules=no --project=test/extensions -e 'using Test, Clapeyron, Unitful; include("test/methods/mass_fractions.jl")'` (2 tests passed)
- `Aqua.test_ambiguities(Clapeyron)`
- `Aqua.test_undefined_exports(Clapeyron)`
- `git diff --check upstream/master...HEAD`

I deferred the full multi-platform test suite to CI.

Fixes #625


longemen3000 (approved):

longemen3000:
@Sush1090, Clapeyron uses a molar basis for everything, so if one passes a mass fraction to a clapeyron function, what would you expect to be the basis?
- if we set the basis of the mass fractions to 1 mol, then the results of MassFractions(w) and the corresponding molar fractions will be the same
- if we set the basis of mass fractions to 1 kg, the results of MassFractions(w) will be different than the results of of a molar fraction, but consistent by passing one kg as an input.

the code solution is easy, just adding an optional argument on the `MolFractions` struct, but the default is important.

Sush1090:
We should keep mol fraction as basis I think.

should there be an assertion that sum of mass and mol fraction should be 1? Else the name might be a bit misleading

PR #638, opened 2026-09-16
Add excess properties for activity models

se-schmitt:
This PR adds dispatches for excess properties of activity models.

---

Addtionally, there might have been a bug for entropy/helmholtz energy of EOS models. That's why I also changed an existing test.

Here a brief formula summary of the changes for excess properties for full EOS:
 
- `helmholtz_energy`: No method existed, so it fell through to `mixing` and returned the mixing property.
  
  New:

  $$a^{E} = a(p,T,\mathbf{z}) - \sum_i z_i \left[ a_{0,i}(p,T) + RT \ln x_i \right]$$

- `entropy`

  old vs new:

```math
  s^{E}_{\mathrm{old}} = s^{\mathrm{res}}(v,T,\mathbf{z}) - \sum_i z_i\, s^{\mathrm{res}}_{0,i}(v_{0,i},T)
```

```math
  s^{E}_{\mathrm{new}} = s(p,T,\mathbf{z}) - \sum_i z_i \left[ s_{0,i}(p,T) - R \ln x_i \right]
```

- `gibbs_energy`: Formula unchanged

Now satisfied, previously not:

$$s^{E} = -\left(\frac{\partial g^{E}}{\partial T}\right)_{p,\mathbf{z}}, \qquad a^{E} = g^{E} - p\,v^{E}$$


se-schmitt:
@longemen3000 Weird erorrs, different versions in different OS/Julia versions... are they related? (I would guess/hope not)

longemen3000 (approved):

longemen3000:
Test errors are unrelated, you catched me in the middle of a refactor of UNIFAC models (got to reduce 1 allocation!) that should be fixed on master.

